PR #9, opened 2022-06-03
feat: add dtmf support

drymarev:


sample/src/main/kotlin/com/pexip/sdk/sample/conference/ConferenceAction.kt
```diff
@@ -5,10 +5,27 @@ import com.pexip.sdk.conference.PresentationStartConferenceEvent
...
+    override fun Updater.apply() {
+        state = state.copy(showingDtmf = !state.showingDtmf)
+    }
+}
+
+class OnDtmfOutput(private val output: DtmfOutput) : ConferenceAction() {
+
+    override fun Updater.apply() {
+        when (output) {
+            DtmfOutput.Back -> state = state.copy(showingDtmf = false)
```
tpettersen, line 24: object DtmfOutput is sufficient, unless you have other plans for the output? And if you want to keep it, "close" is a better word as it is a dialog. Nitpick ftw on a tuesday morning 😅

tpettersen (approved):
lgtm :) 1 minor suggestion only

sample/src/main/kotlin/com/pexip/sdk/sample/conference/ConferenceAction.kt
drymarev: Prefer to be explicit, just `DtmfOutput` doesn't quite explain what's going on.

PR #29, opened 2022-09-01
refactor: provide ViewEnvironment using Dagger

drymarau:
- refactor: provide ViewEnvironment using Dagger
- refactor: extract ComposerWorkflow
- refactor: remove DtmfProps


tpettersen (approved):
I think this is a better approach, and it is contained to the specific module without having to define everything in root 👍 Good job! 

It also opens up for several viewRegistries if we need to handle some stuff differently - like we do with `EmptyPresentationViewFactory`

PR #33, opened 2022-09-30
feat: PermissionsWorkflow

drymarau:
- refactor: WelcomeWorkflow -> DisplayNameWorkflow
- feat: PermissionsWorkflow


sample/src/main/res/values/strings.xml
```diff
@@ -1,8 +1,15 @@
...
 <resources>
     <string name="app_name">"Video SDK"</string>
+    <string name="app_name_full">"Pexip Video SDK"</string>
     <!-- CONFERENCE -->
     <string name="conference_notification_channel_id" translatable="false">conference</string>
     <string name="conference_notification_channel_name">Conference</string>
     <string name="conference_notification_content_title">Conference</string>
+    <!-- PERMISSIONS -->
+    <string name="permissions_title">Grant permissions</string>
+    <string name="permissions_description">Granting permission in the next screen will make sure other participants are able to see and hear you if you’ve enabled your camera and microphone</string>
```
tpettersen, line 11: nitpicky tuesday morning: "next screen" is not quite accurate for the popup that comes

sample/src/main/kotlin/com/pexip/sdk/sample/permissions/PermissionsAction.kt
```diff
@@ -0,0 +1,29 @@
...
+
+typealias PermissionsAction = WorkflowAction<PermissionsProps, PermissionsState, PermissionsOutput>
+
+class OnPermissionsRequestResult(private val result: PermissionsRequestResult) :
+    PermissionsAction() {
+
+    override fun Updater.apply() {
+        val output = when {
+            result.grants.all() -> PermissionsOutput.Next
+            result.rationales.none() && state.rationales.none() -> PermissionsOutput.ApplicationDetailsSettings
```
tpettersen, line 13: This triggers a slightly weird behavior. When seeing the dialog for the first time, pressing outside the dialog, I get sent to ApplicationDetails. I would expect the dialog to just close. That does not happen if I at some point remove 1 setting, gets the popup again and press outside the popup

tpettersen (approved):
It works good and feels better with an explanation up front 👍  I had the tiniest nitpicks that you can do whatever you want with.

sample/src/main/res/values/strings.xml
drymarau: Copied from ascari, and more generic (nothing stopping OEMs from making it an actual "screen" too)

sample/src/main/kotlin/com/pexip/sdk/sample/permissions/PermissionsAction.kt
tpettersen: Does not sound like something worth fixing 😅 Super minor case

PR #34, opened 2022-10-05
feat: PreflightWorkflow

drymarau:
This PR introduces a new start destination in navigation graph - preflight.

It somewhat follows the existing UX of Pexip Infinity Connect apps and will enable call transfer handling in the near future.

### DIFF
--- a/sample/src/main/kotlin/com/pexip/sdk/sample/SampleWorkflow.kt
+++ b/sample/src/main/kotlin/com/pexip/sdk/sample/SampleWorkflow.kt
@@ -33,11 +33,15 @@ class SampleWorkflow @Inject constructor(
 ) : StatefulWorkflow<Unit, SampleState, SampleOutput, Any>() {
 
     override fun initialState(props: Unit, snapshot: Snapshot?): SampleState {
-        val screen = when (Permissions.all(context::isPermissionGranted)) {
-            true -> snapshot?.toParcelable() ?: SampleDestination.Preflight
-            else -> SampleDestination.Permissions
-        }
-        return SampleState(screen)
+        val allGranted = Permissions.all(context::isPermissionGranted)
+        return SampleState(
+            destination = when (allGranted) {
+                true -> snapshot?.toParcelable() ?: SampleDestination.Preflight
+                else -> SampleDestination.Permissions
+            },
+            createCameraVideoTrackCount = if (allGranted) 1u else 0u,
+            createMicrophoneAudioTrackCount = if (allGranted) 1u else 0u
+        )
     }
 
     override fun snapshotState(state: SampleState): Snapshot = state.destination.toSnapshot()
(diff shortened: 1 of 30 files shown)


sample/src/main/kotlin/com/pexip/sdk/sample/preflight/PreflightScreen.kt
```diff
@@ -0,0 +1,73 @@
...
+                }
+            }
+        }
+    }
+    if (rendering.childRendering != null) {
+        WorkflowRendering(
+            rendering = rendering.childRendering,
+            viewEnvironment = environment,
+        )
+    }
```
tpettersen, line 72: It is somewhat annoying that I lose state pressing outside the popups by accident. E.g: I introduce myself, write the meeting uri + host and the touches the screen outside of the popup when pin challenge is presented. Then I have to start all over again. That is a bit unfortunate

sample/src/main/kotlin/com/pexip/sdk/sample/SampleWorkflow.kt
```diff
@@ -3,79 +3,130 @@ package com.pexip.sdk.sample
...
+
+    private fun RenderContext.createMicrophoneAudioTrackSideEffect(screen: SampleDestination) {
+        if (screen != SampleDestination.Permissions) runningSideEffect("createMicrophoneAudioTrackSideEffect") {
+            val track = localAudioTrackFactory.createLocalAudioTrack()
+            val action = OnMicrophoneAudioTrackChange(track)
+            actionSink.send(action)
+        }
+    }
+
+    private fun RenderContext.cameraVideoTrackSideEffect(track: CameraVideoTrack?) {
```
tpettersen, line 106: There is a lifecycle issue with the camera:
- Open preflight
- Put the app in the background
- Open the camera app on your phone
- Open Video SDK app again

The camera is now gone, and you will not get it back when entering a conference even. The app needs to restart to get back to normal.

tpettersen (changes requested):
The UI looks sharp 👍 Found a couple of issues

sample/src/main/kotlin/com/pexip/sdk/sample/SampleWorkflow.kt
```diff
@@ -80,8 +84,8 @@ class SampleWorkflow @Inject constructor(
         }
     }
 
-    private fun RenderContext.createCameraVideoTrackSideEffect(screen: SampleDestination) {
-        if (screen != SampleDestination.Permissions) runningSideEffect("createCameraVideoTrackSideEffect") {
+    private fun RenderContext.createCameraVideoTrackSideEffect(count: UInt) {
+        if (count > 0u) runningSideEffect("createCameraVideoTrackSideEffect($count)") {
```
tpettersen, line 88: This _works_, but I must admit that I am not to fan of counters here. 😅 

Just to have the discussion, what if you added `onCameraConnected(deviceName: String)` in `CameraVideoTrack.Callback`. Then you could store that deviceName in the state. 

`OnCameraDisconnected` sets the track to null, but you would still have the `deviceName`. So using `createCameraVideoTrack(deviceName, callback)` when the app is in the background would throw `IllegalStateException` if it is not available, and would not produce more renderings. The next time the app opens again, this should now work

drymarau: I don't think it's a good approach as the `deviceName` may not necessarily be static and overall feels way more complex to track application lifecycle.

drymarau: > Huh, could the device name of the current camera actually change runtime? 🤔

If it's an external camera it's definitely possible, there are no guarantees about `deviceName` not changing.

Use cases differ and I don't think it's feasible to provide Android lifecycle integration at all - it's fairly easy, if required, to build one yourself.

drymarau: As for in-call, general advice is to not open camera app while in a conference call.

tpettersen: ok 👌

PR #38, opened 2023-01-04
feat: use spotless

drymarau:


gradle/libs.versions.toml
```diff
@@ -11,9 +11,10 @@ cachefix = "2.5.7"
 dagger = "2.44"
 kotlin = "1.7.20"
 kotlinx-coroutines = "1.6.4"
-ktlint = "11.0.0"
+ktlint = "0.48.1"
```
tpettersen, line 14: Do we need to keep this? It is not in use anywhere as far as I can see

tpettersen (approved):

gradle/libs.versions.toml
drymarau: `ktlint(libs.versions.ktlint.get())` in `kotlin { }` and `kotlinGradle { }` blocks.

tpettersen: ah. That is what I get for searching for `libs.ktlint` 😅

PR #54, opened 2023-02-27
feat: introduce Messenger

drymarau:
This interface provides a way to both send and receive `Message`s and is
a direct replacement for `Conference.message()` and
`MessageReceivedConferenceEvent`.

It also supports direct messaging via `Messenger.send()` that accepts a
participant ID.

`MessageNotSendException` should enable clients to provide retries since
it includes both an instance of a `Message` that would've been received
if the call was successful and an optional `cause`.

Note that `MessageListener` will not be invoked for messages that are
sent by the caller (which differs from previous version). Clients are
encouraged to inject an instance of the `Message` obtained from
`SendCallback.onSuccess()` to the list they maintain manually. Previous
behavior might be undesirable for applications that would like to
utilize message types other than "text/plain" (i.e. non-chat use cases).


sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/MessengerImpl.kt
```diff
@@ -0,0 +1,121 @@
...
+import com.pexip.sdk.api.infinity.TokenStore
+import com.pexip.sdk.conference.Message
+import com.pexip.sdk.conference.MessageListener
+import com.pexip.sdk.conference.MessageNotSentException
+import com.pexip.sdk.conference.Messenger
+import com.pexip.sdk.conference.SendCallback
+import java.util.UUID
+import java.util.concurrent.CopyOnWriteArraySet
+
+internal class MessengerImpl(
```
vadymmarkov, line 35: Is the plan to add another implementation of `Messenger` interface for sending messages via WebRTC data channel?

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/MessageNotSentException.kt
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Thrown to indicate that the [msg] was not sent.
+ *
+ * @property msg an instance of [Message] that was not sent
+ * @property cause a cause of this exception
+ */
+public class MessageNotSentException @JvmOverloads constructor(
+    public val msg: Message,
+    cause: Throwable? = null,
+) : RuntimeException("Failed to sent the message.", cause)
```
tpettersen, line 27:
```suggestion
) : RuntimeException("Failed to send the message.", cause)
```

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Messenger.kt
```diff
@@ -0,0 +1,57 @@
...
+     * @param callback a callback to be invoked upon completion
+     */
+    public fun send(participantId: UUID, type: String, payload: String, callback: SendCallback)
+
+    /**
+     * Registers a [MessageListener].
+     *
+     * @param listener a message listener
+     */
+    public fun registerMessengerListener(listener: MessageListener)
```
tpettersen, line 49: Perhaps better naming to have `registerMessageListener` or `listener: MessengerListener` to align them, like we do for conference, registration and so on?

CHANGELOG.md
```diff
@@ -13,11 +13,13 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 - `WebRtcMediaConnectionFactory.createLocalVideoTrack` that accepts `VideoCapturer`
 - `ParticipantStep.message` that allows sending direct messages
 - `MessageReceivedEvent.direct` that indicates whether it's direct message
+- `Messenger` that provides facilities to send and receive `Message`s
 
 ### Changed
 
 - WebRTC to M110
 - **BREAKING**: `MessageRequest` no longer has a default `type` set
+- Deprecated `Conference.message(String)`
```
tpettersen, line 22: Perhaps also mention the part you added in the PR description

> `MessageListener` will not be invoked for messages that are sent by the caller

tpettersen (approved):
Looks good 👍 Added a couple of suggestions

sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/MessengerImpl.kt
drymarau: Either another implementation or move the transport to `MediaConnectionSignaling`.

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Messenger.kt
drymarau: Yup, forgot to rename these methods after rename of the listener.

PR #57, opened 2023-03-02
feat: bump gradle to 8.0.1

drymarau:


tpettersen (approved):
builds and runs :) Not wanting to upgrade Gradle Build Plugin to 8.1 at the same time?

drymarau:
> Not wanting to upgrade Gradle Build Plugin to 8.1 at the same time?

Can you elaborate a bit what plugin exactly do you mean?

tpettersen:
> > Not wanting to upgrade Gradle Build Plugin to 8.1 at the same time?
> 
> Can you elaborate a bit what plugin exactly do you mean?

The Android one for the Sample app: https://developer.android.com/studio/releases/gradle-plugin#updating-gradle. But when I read on the release notes I can only see `7.4.2` 🤔 Weird

drymarau:
> The Android one for the Sample app: https://developer.android.com/studio/releases/gradle-plugin#updating-gradle. But when I read on the release notes I can only see `7.4.2` 🤔 Weird

8.1 would correspond to Giraffe I think, which is still in Canary. 8.0 (Flamingo) is in Beta, so still a bit early.

PR #60, opened 2023-03-08
feat: bump kotlin to 1.8.10

drymarau:
https://github.com/JetBrains/kotlin/releases/tag/v1.8.10
https://github.com/Kotlin/dokka/releases/tag/v1.8.10
https://github.com/Kotlin/kotlinx.serialization/releases/tag/v1.5.0-RC
https://github.com/Kotlin/kotlinx.serialization/releases/tag/v1.5.0
https://github.com/Kotlin/binary-compatibility-validator/releases/tag/0.13.0

ProGuard rules are now bundled with `kotlinx.serialization` so we can safely remove our rules.


gradle/libs.versions.toml
```diff
@@ -13,10 +13,10 @@ androidx-startup = "1.1.1"
 accompanist = "0.28.0"
 cachefix = "2.6.5"
 dagger = "2.44.2"
-dokka = "1.7.20"
-kotlin = "1.7.20"
+dokka = "1.8.10"
+kotlin = "1.8.10"
 kotlinx-coroutines = "1.6.4"
```
tpettersen, line 18: I see that `1.7.0-Beta` is updated to use 1.8.10, but that does not seem to make any difference for us when testing.

tpettersen (approved):
👍

gradle/libs.versions.toml
drymarau: Would rather wait until full release.

PR #62, opened 2023-03-08
ci: upload snyk's output

drymarau:


tpettersen (approved):
Works well testing it locally 👌

PR #63, opened 2023-03-16
feat: add dependabot

tpettersen:


tpettersen:
Did not setup commit signing. So had to open a new PR for it

PR #85, opened 2023-04-06
fix(sample): crash when calling FocusRequester.freeFocus()

drymarau:
It's only necessary to invoke this method to match previous
`FocusRequester.captureFocus()`. Since we don't do that, it crashes the
app.


tpettersen (approved):
It did never crash in my end, but the alias dialog got stuck while the camera was still moving around. Now it all works smoothly

drymarau:
I've encountered it on the emulator, so it might be happening just there.

PR #86, opened 2023-04-06
fix: insert black frame when VideoCapturer stops

drymarau:
When a client stops capturing, last sent frame will be left lingering in
the conference. While this is not an issue if this change is signalled
via REST API, it affects users in gateway calls where removing from mix
is not supported.

This commit inserts a black frame when `onCapturerStopped()` in invoked.

### DIFF
--- a/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcLocalVideoTrack.kt
+++ b/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcLocalVideoTrack.kt
@@ -64,7 +64,7 @@ internal open class WebRtcLocalVideoTrack(
 
         override fun onCapturerStopped() {
             val duration = TimeUnit.MILLISECONDS.toNanos(100)
-            repeat(2) {
+            repeat(4) {
                 val buffer = createBlackBuffer(width, height)
                 val frame = VideoFrame(buffer, rotation, timestampNs + (it + 1) * duration)
                 videoSource.capturerObserver.onFrameCaptured(frame)
--- a/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcLocalVideoTrack.kt
+++ b/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcLocalVideoTrack.kt
@@ -16,7 +16,6 @@
 package com.pexip.sdk.media.webrtc.internal
 
 import android.content.Context
-import android.util.Rational
 import com.pexip.sdk.media.LocalMediaTrack
 import com.pexip.sdk.media.LocalVideoTrack
 import com.pexip.sdk.media.QualityProfile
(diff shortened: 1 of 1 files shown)


vadymmarkov:
In the Swift SDK we render empty frame when video capturer stops by calling `renderFrame(nil)` on `RTCVideoRenderer`, which I believe does the same thing. But I guess you don't have `RTCVideoRenderer` on Android.

drymarau:
Does it affect the remote end as well or just the local?

vadymmarkov:
You're right, I misread the issue. It affects just the local. 
But shouldn't it be enough to set `isEnabled` to false on `RTCVideoTrack` before you stop video capturer?

drymarau:
@tpettersen, I'm able to reproduce this issue. Can you check if passing `width` and `height` directly instead of frame calculation works?

tpettersen:
I figured out that there is something off when creating the black buffer. If I change the code to do this `val buffer = createBlackBuffer(width, height)` it works.

tpettersen:
> @tpettersen, I'm able to reproduce this issue. Can you check if passing `width` and `height` directly instead of frame calculation works?

..yeah 😅 you beat me to it with seconds

drymarau:
> I figured out that there is something off when creating the black buffer. If I change the code to do this `val buffer = createBlackBuffer(width, height)` it works.

It appears that resulting frame's size when it's derived from `Rational` is not very welcome by the encoder, hence the crash.

I've tried to pass a 256x144 frame and it works on S23. Maybe something static like 640x480 would do the trick without keeping aspect ratio.

vadymmarkov:
> I tried that approach but couldn't get it to work reliably - it appears at least one frame has to be sent after the track has been disabled for remote end to see the black frame.

On iOS we add a small delay between disabling the track and stopping the capturer, which is arguably not the best way to solve this, but proved to be working in our apps. 

There is also the case when the app crashes during the call or is closed by user, which would cause the same problem. But I don't think there is anything we can do about it on the client.

tpettersen (changes requested):
This does not work for me I am afraid. The frame is still lingering. I think you need to allocate a buffer with the correct size and a frame with the correct rotation. 

When testing I pass width/height for my device and it works as intended in landscape only. If I also set the correct rotation it works both portrait and landscape. 

```
val buffer = createBlackBuffer(width, height)
val frame = VideoFrame(buffer, rotation, 0)
```

drymarau:
> When testing I pass width/height for my device and it works as intended in landscape only. If I also set the correct rotation it works both portrait and landscape.

Doesn't work for me reliably - frames with `rotation=0` stay while `rotation=180` are replaced with a black one as expected.

tpettersen:
> > When testing I pass width/height for my device and it works as intended in landscape only. If I also set the correct rotation it works both portrait and landscape.
> 
> Doesn't work for me reliably - frames with `rotation=0` stay while `rotation=180` are replaced with a black one as expected.

True. Works for me with `180` and `270`, but not `0` 🤔

drymarau:
> True. Works for me with `180` and `270`, but not `0` 🤔

So sending at least 2 frames with correctly set `timestampNs` regardless of frame size and rotation seems to work, though I need to verify it for other codecs.

drymarau:
Guess I'll bump it to 4 frames 🙂

PR #90, opened 2023-04-14
feat: build with jdk11

drymarau:
Some of our dependencies (primarily WebRTC) have switched to JDK11, so we also start building with them.

Note that the plugins (and CI) are JDK17 in preparation for AGP 8.0.

### DIFF
--- a/.idea/kotlinc.xml
+++ b/.idea/kotlinc.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="KotlinJpsPluginSettings">
+    <option name="version" value="1.8.20-RC" />
+  </component>
+</project>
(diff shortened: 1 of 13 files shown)


.idea/kotlinc.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="KotlinJpsPluginSettings">
+    <option name="version" value="1.8.20-RC" />
```
tpettersen, line 4: Sure about us using Release Candidates? I would go for 1.8.10 and do the upgrade in #84

drymarau: We don't explicitly depend on it, it appears that it's coming from one of the compiler plugins (`kotlinNativeCompilerPluginClasspath`). I think we can safely merge since there's #84 anyway.

tpettersen (approved):

PR #98, opened 2023-04-24
feat: bump gradle to 8.0.2, improve CI

drymarau:
- feat: bump gradle to 8.0.2
- ci: share GRADLE_OPTS
- ci: use gradle-build-action


.github/workflows/pull_request.yml
```diff
@@ -22,8 +22,8 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '17'
-          cache: 'gradle'
 
+      - uses: gradle/gradle-build-action@v2
```
tpettersen, line 26: In my mind it sounds more correct that `main` writes to the cache and the `pull requests` only reads, to not mix the cache between PRs. What is the reasoning behind this approach - or rather; what am I missing?

drymarau: By default this action will only write caches on default branches (e.g. `main`/`master`). It will be in read-only mode in PRs.

Explicit `cache-read-only: true` on `main` is only for Snyk-related tasks.

See actions [README](https://github.com/gradle/gradle-build-action#using-the-caches-read-only) for more info.

tpettersen: Ah, I see 👌

PR #122, opened 2023-06-01
chore: warn on outdated infinity deployment

drymarau:


sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/InfinityConference.kt
```diff
@@ -113,6 +114,16 @@ public class InfinityConference private constructor(
             step: InfinityService.ConferenceStep,
             response: RequestTokenResponse,
         ): InfinityConference {
+            if (response.version.versionId < "29") {
```
tpettersen, line 117: Out of curiosity, how low a version before it becomes a problem? If random stuff starts to happen or the SDK behaves really bad, we should rather throw here I think.

drymarau: We'll throw in one of the upcoming releases early next year. 

As far as version goes, I don't have a concrete value.

tpettersen (approved):

PR #125, opened 2023-06-09
misc: add issue templates

drymarau:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,33 @@
...
+    attributes:
+      label: 'Expected behavior'
+      description: 'A clear and concise description of what you expected to happen.'
+    validations:
+      required: true
+  - type: input
+    attributes:
+      label: 'Version'
+      description: 'What library version are you using?'
+      placeholder: '1.0.0'
```
tpettersen, line 27: We could for the sake of ease for the user set this to the current released version of the SDK every time we push a new version

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,26 @@
...
+      label: 'Describe the solution you would like'
+      description: 'A clear and concise description of what you want to happen.'
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: 'Describe alternatives you have considered'
+      description: 'A clear and concise description of any alternative solutions or features you have considered.'
+    validations:
+      required: true
```
tpettersen, line 22: Does alternatives have to be required? If I come up with "I would like to have DTMF", there is not quite many alternatives that would be relevant

tpettersen (approved):
👍 I added a couple of suggestions

.github/ISSUE_TEMPLATE/bug_report.yml
drymarau: 👍

PR #132, opened 2023-06-26
refactor(plugins): merge :android and :kotlin

drymarau:
Keeping them separate doesn't give us any advantages


plugins/convention/src/main/kotlin/com/pexip/sdk/KotlinAndroidApplicationPlugin.kt
```diff
@@ -0,0 +1,41 @@
...
+            apply(AndroidApplicationPlugin::class)
+            apply("org.jetbrains.kotlin.android")
+        }
+        dependencies {
+            "testImplementation"(kotlin("test-junit"))
+            "androidTestImplementation"(kotlin("test-junit"))
+        }
+        tasks.withType<Test>().configureEach {
+            systemProperty("kotlinx.coroutines.stacktrace.recovery", false)
+        }
```
tpettersen, line 39: Would it make sense to add this to a TestPlugin, along with `assertK`, and reuse it in `KotlinAndroidLibraryPlugin`?

tpettersen (approved):
Looks good 👌 Small suggestion only

plugins/convention/src/main/kotlin/com/pexip/sdk/KotlinAndroidApplicationPlugin.kt
drymarau: I think we it's solvable by introducing a `KotlinAndroidPlugin` (similar to what we have internally) that setups this plus a few other Kotlin-related things.

PR #133, opened 2023-06-29
chore: remove dangling directory

drymarau:
Forgot to remove it previously, but it was already excluded from
`settings.gradle.kts`


vadymmarkov (approved):
Always nice to remove some code 😄

PR #136, opened 2023-07-07
refactor(media): defer ack

drymarau:
This should help with the _broken camera_ icon in gateway calls by
letting the `MediaConnection` have answer set earlier.


vadymmarkov (approved):
👍🏻

PR #142, opened 2023-07-10
refactor: merge -coroutines with base modules

drymarau:
This PR merges all currently available coroutines extensions with their respective base modules. 

It should enable broader usage of coroutines, as well as open a road for Kotlin Multiplatform in the future.

The `-coroutines` modules will continue to be published for some time to provide a clean migration path.


tpettersen (approved):
Looks alright :) Would it make sense to prefix the POM-description for the `-coroutines` packages with a "deprecated: use sdk-x"?

drymarau:
> Would it make sense to prefix the POM-description for the `-coroutines` packages with a "deprecated: use sdk-x"?

Given that it's not visible anywhere apart from Maven Central I'd say no. `CHANGELOG.md` should be fine I think.

PR #143, opened 2023-07-10
refactor(media): preparation for perfect negotiation

drymarau:
This PR adds a handful of techniques for perfect negotiation. The full guide can be found over at https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Perfect_negotiation

### DIFF
--- a/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
+++ b/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
@@ -310,6 +310,13 @@ internal class WebRtcMediaConnection(
         presentationRemoteVideoTrackListeners -= listener
     }
 
+    override fun onStandardizedIceConnectionChange(newState: PeerConnection.IceConnectionState) {
+        if (newState != PeerConnection.IceConnectionState.FAILED) return
+        workerExecutor.maybeExecuteUnlessDisposed {
+            connection.restartIce()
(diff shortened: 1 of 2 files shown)


sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
```diff
@@ -316,15 +310,22 @@ internal class WebRtcMediaConnection(
...
+            connection.restartIce()
+        }
+    }
+
     override fun onIceCandidate(candidate: IceCandidate) {
         onCandidate(candidate.sdp, candidate.sdpMid)
     }
 
     override fun onRenegotiationNeeded() {
         // Skip the first call to onRenegotiationNeeded() since it's called right after
```
vadymmarkov, line 325: Not sure if you need this, but I also have a check here that connection state is not `closed` or `disconnected`. If I remember correctly, that was added to avoid unnecessary negotiation calls after `MediaConnection.stop`.

vadymmarkov (approved):
Looks good 👍🏻

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
drymarau: Good point, although it should be handled already via `Executor. maybeExecuteUnlessDisposed()`.

PR #148, opened 2023-07-13
refactor(media, conference)!: make MediaConnectionSignaling a property

drymarau:
The move is motivated by the following:
1. Modifying `MediaConnectionSignaling` also affects `Conference` API, which often leads to tests not compiling (i.e. if a new method is added)
2. It may prove useful to be able to create new instances of `MediaConnectionSignaling` upon invocation (for scoping purposes)

This is a breaking change and consumers will be unable to compile after the update. The fix is to change `MediaConnectionConfig.Builder(conference)` to `MediaConnectionConfig.Builder(conference.signaling)`.


vadymmarkov (approved):
This is also how it's implemented in Swift SDK. The only difference is that the property is called `signalingChannel` for some reason 🙂

PR #152, opened 2023-07-18
feat(media): limit max framerate for encoders

drymarau:
This may help with some devices having trouble when the encoder could be
configured above 30 fps for 1080p.


sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/Util.kt
```diff
@@ -144,3 +168,18 @@ internal fun RtpTransceiver.setTrack(track: LocalMediaTrack?) {
...
+) = RtpTransceiver.RtpTransceiverInit(direction, streamIds, sendEncodings)
+
+internal fun Encoding(
+    rid: String? = null,
+    active: Boolean = true,
+    scaleResolutionDownBy: Double? = null,
+    block: Encoding.() -> Unit = { },
+) = Encoding(rid, active, scaleResolutionDownBy).apply(block)
+
+internal const val MAX_FRAMERATE = 30
```
tpettersen, line 185:
```suggestion
internal const val MAX_FRAME_RATE = 30
```

tpettersen (approved):

tpettersen:
Just a question. If my devices supports e.g. 4k 60FPS, would it not be logical to expose an override in the API for max frame rate? Or is this a limitation on Infinity?

drymarau:
> Just a question. If my devices supports e.g. 4k 60FPS, would it not be logical to expose an override in the API for max frame rate? Or is this a limitation on Infinity?

Infinity itself is limited to 1080p30 (see [docs](https://docs.pexip.com/admin/call_quality.htm#max_quality)), but you could technically go higher than that in a direct media call provided your device supports it.

We can and definitely should make it configurable, maybe with a soft limit of 30fps for transcoded calls.

This particular PR mostly addresses an issue with one specific device we have where it's unable to handle 1080p60, but the framework attempts to configure the encoder to just that, which in the end results in failures when more than one video stream is active.

PR #153, opened 2023-07-20
refactor(media): replace executors with coroutines

drymarau:
- refactor(media): use coroutines in WebRtcMediaConnection
- refactor(media): keep VideoTrack in a StateFlow
- refactor(media): remove synchronized usage


tpettersen (approved):
Works good :)

PR #167, opened 2023-08-04
fix(media): pin MediaConnectionFactory-produced objects to a single thread

drymarau:
Note that `PeerConnectionWrapper` now assumes that it will be used from
a single thread and doesn't make any attempts to protect a handful of
properties from concurrent access.


tpettersen (approved):
Seems alot more stable

PR #171, opened 2023-08-07
refactor(media): share a parent Job between children of MediaConnectionFactory

drymarau:
Instead of making sure all the children complete before `dispose()` is called, we'll instead cancel their parent `Job`.

### DIFF
--- a/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/WebRtcMediaConnectionFactory.kt
+++ b/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/WebRtcMediaConnectionFactory.kt
@@ -36,17 +36,17 @@ import com.pexip.sdk.media.webrtc.internal.WebRtcLocalAudioTrack
 import com.pexip.sdk.media.webrtc.internal.WebRtcLocalVideoTrack
 import com.pexip.sdk.media.webrtc.internal.WebRtcMediaConnection
 import com.pexip.sdk.media.webrtc.internal.from
-import kotlinx.coroutines.CompletableJob
+import com.pexip.sdk.media.webrtc.internal.maybeExecute
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.asExecutor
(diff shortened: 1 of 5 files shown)


sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/WebRtcMediaConnectionFactory.kt
```diff
@@ -228,30 +221,18 @@ public class WebRtcMediaConnectionFactory private constructor(
...
-        }
-        if (disposed.compareAndSet(false, true)) {
-            workerDispatcher.executor.execute {
-                factory.dispose()
-                audioDeviceModule.release()
-            }
-            workerDispatcher.close()
-        } else {
-            throw IllegalStateException("WebRtcMediaConnectionFactory has been disposed!")
+        runBlocking { job.cancelAndJoin() }
```
tpettersen, line 230: This can throw `CancellationException`. Should it be handled?

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/WebRtcMediaConnectionFactory.kt
```diff
@@ -136,9 +132,8 @@ public class WebRtcMediaConnectionFactory private constructor(
             context = applicationContext,
             audioSource = audioSource,
             audioTrack = audioTrack,
-            workerDispatcher = workerDispatcher,
+            scope = CoroutineScope(),
```
tpettersen, line 135: I don't quite follow as to why we need to create a new `CoroutineScope` 4 times. Would it not suffice to have it as a variable?

tpettersen (approved):
Seems to work alright. Left a couple of questions

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/WebRtcMediaConnectionFactory.kt
drymarau: As far as I understand, it can only throw if the parent `CoroutineScope` (in this case provided by `runBlocking`) is cancelled, which shouldn't happen.

`CancellationException` in general shouldn't be handled, just re-thrown, as to not break structured concurrency.

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/WebRtcMediaConnectionFactory.kt
drymarau: No, as all instances created by the factory have a different "lifecycle". If we have it as a single property we have to come up with a way to create a child scope that should be cancellable on its own, without affecting the parent scope.

PR #172, opened 2023-08-07
refactor(api): introduce TokenStore.refreshTokenIn()

drymarau:


sdk-api-infinity/src/main/kotlin/com/pexip/sdk/api/infinity/TokenStore.kt
```diff
@@ -40,5 +49,38 @@ public interface TokenStore {
...
+         */
+        public inline fun TokenStore.refreshTokenIn(
+            scope: CoroutineScope,
+            crossinline refreshToken: suspend (Token) -> Token,
+            crossinline releaseToken: suspend (Token) -> Unit,
+            crossinline onFailure: (t: Throwable) -> Unit,
+        ): Job = scope.launch {
+            try {
+                while (isActive) {
+                    val response = refreshToken(get())
```
tpettersen, line 71: Would be nice to introduce a bit more robustness in another PR. If the refreshToken fails just once for some reason, a `FailureConferenceEvent` or `FailureRegistrationEvent` is thrown, and the refresh stops. Exponential backoff ftw :)

tpettersen (approved):
Way easier to read than the previous `TokenRefresher` 🙌

sdk-api-infinity/src/main/kotlin/com/pexip/sdk/api/infinity/TokenStore.kt
drymarau: True, there is some room for improvement here.

PR #174, opened 2023-08-08
feat(api): add RequestTokenRequest.registrationToken

drymarau:
To exchange a registration token for conference token we must use a this header instead of `token`.


tpettersen (approved):
Would have been nice to find this API documented elsewhere than a PDF on Slack from 2018...

PR #178, opened 2023-08-21
fix(media): moved keepScreenOn outside of renderer

tpettersen:
The `keepScreenOn` inside AndroidView is not working for me, neither in the SampleApp or in the apps using the SDK. So I moved it to a disposableEffect instead, which seems to work fine.


tpettersen:
> This doesn't seem quite right - `LocalView` might be referenced from another place where the value would be overridden.

Hm. Any other ideas then? It seem to work fine whenever I use the `VideoTrackRenderer`

tpettersen:
> Maybe setting it in `update` will work. I'd also include it as a default parameter in the function.

Seems to work alright 👍

drymarau (approved):
Some small changes necessary, please see comments.

PR #189, opened 2023-08-28
feat: bump compileSdk to 34

drymarau:
`targetSdk` will be bumped a bit later (only affects the sample app) as it requires a few `Service`-related changes.


tpettersen (approved):
Builds and runs fine 👍

PR #190, opened 2023-08-28
fix(api): registration_token as JSON property

drymarau:
Due to historical reasons `registration_token` cannot be used as a header and must be provided in the body instead.


tpettersen (approved):
Sadly, it looks ok

PR #198, opened 2023-09-01
feat(api): add a way to navigate to a previous *Step

drymarau:


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- **BREAKING**: `InfinityService.RequestBuilder.infinityService`,`InfinityService.ConferenceStep.requestBuilder`,
```
tpettersen, line 12: Is the change actually breaking anything? The external apis is not changes as far as I can see

tpettersen (approved):

CHANGELOG.md
drymarau: It is a breaking change for clients that extend these interfaces. But I do agree that it's a rare case and we can drop it.

PR #201, opened 2023-09-05
feat(conference): introduce Referer

drymarau:
The idea here is to provide a streamlined way to transfer calls.

### DIFF
--- a/sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Referer.kt
+++ b/sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Referer.kt
@@ -28,6 +28,7 @@ public interface Referer {
      *
      * @param event a [ReferConferenceEvent] that contains information about target conference
      * @return a new [Conference]
+     * @throws [ReferException] if [refer] did not succeed
      */
     public suspend fun refer(event: ReferConferenceEvent): Conference
 }


sample/src/main/kotlin/com/pexip/sdk/sample/conference/ConferenceWorkflow.kt
```diff
@@ -196,7 +186,7 @@ class ConferenceWorkflow @Inject constructor(
     private fun RenderContext.leaveSideEffect(
         renderProps: ConferenceProps,
         renderState: ConferenceState,
-    ) = runningSideEffect("${renderState.conference}Leave") {
+    ) = runningSideEffect("${renderProps.conference}Leave") {
```
tpettersen, line 189:
```suggestion
    ) = runningSideEffect("leaveSideEffect(${renderProps.conference})") {
```

sample/src/main/kotlin/com/pexip/sdk/sample/pinchallenge/PinChallengeWorkflow.kt
```diff
@@ -65,11 +66,9 @@ class PinChallengeWorkflow @Inject constructor(private val store: SettingsStore)
             val action = try {
                 val displayName = store.getDisplayName().first()
                 val request = RequestTokenRequest(displayName = displayName)
-                val response = props.builder
-                    .conference(props.conferenceAlias)
-                    .requestToken(request, pinToSubmit)
-                    .await()
-                OnResponse(response)
+                val response = props.step.requestToken(request, pinToSubmit).await()
+                val conference = InfinityConference.create(props.step, response)
```
tpettersen, line 70: Semantically I think it is a bit off that `PinChallengeWorkflow` and `PinRequirementWorkflow` creates the conference. It could maybe have been better with a `CreateConferenceWorkflow` which does the node check and creates the builder, and then pass the builder/step to PinRequirement + PinChallenge which requests token and outputs the result. But no biggie at all - semantics only :)

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Referer.kt
```diff
@@ -0,0 +1,33 @@
...
+public interface Referer {
+
+    /**
+     * Transfers the call using information specified in the [event].
+     *
+     * Note that upon successful call transfer the current [Conference] will still be active and
+     * must be disposed of manually.
+     *
+     * @param event a [ReferConferenceEvent] that contains information about target conference
+     * @return a new [Conference]
```
tpettersen, line 30:
```suggestion
     * @return a new [Conference]     
     * @throws [ReferException] on failures
```

tpettersen (approved):
Works good! Nice to not having to drag around stuff

PR #221, opened 2023-10-06
feat(media): add MediaConnectionConfig.continualGathering

### LINKED ISSUE #208: Need for end-of-candidates signal in SDK
### Is your feature request related to a problem? Please describe.

If ICE is unable to gather local candidates, no `new_candidate` messages will be sent towards Pexip Infinity. However, in order to glean peer-reflexive candidates from incoming traffic, Infinity needs to have first received an end-of-candidates (empty ICE candidate) signal from the client.

### Describe the solution you would like

Although we are unsure whether a client would attempt to send media in this situation, the client sending an end-of-candidates would put Infinity in the correct state to process any incoming media.

drymarau:
Resolves #208

### DIFF
--- a/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
+++ b/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
@@ -269,20 +269,35 @@ internal class WebRtcMediaConnection(
         .onEach { setLocalDescription() }
         .launchIn(this)
 
-    private fun CoroutineScope.launchOnIceCandidate() = wrapper.event
-        .filterIsInstance<Event.OnIceCandidate>()
-        .onEach {
-            val (ufrag, pwd) = wrapper.getIceCredentials(it.candidate) ?: return@onEach
+    private fun CoroutineScope.launchOnIceCandidate() = launch {
+        fun onCandidate(sdp: String, mid: String, credentials: IceCredentials) = launch {
             runCatching {
                 config.signaling.onCandidate(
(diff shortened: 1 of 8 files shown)


sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
```diff
@@ -262,20 +262,35 @@ internal class WebRtcMediaConnection(
...
+            when (it) {
+                is Event.OnIceCandidate -> {
+                    val sdp = it.candidate.sdp?.takeIf(String::isNotBlank) ?: return@collect
+                    val mid = it.candidate.sdpMid ?: return@collect
+                    val credentials = wrapper.getIceCredentials(mid) ?: return@collect
+                    onCandidate(sdp = sdp, mid = mid, credentials = credentials)
+                }
+                is Event.OnIceGatheringChange -> {
+                    if (it.state != PeerConnection.IceGatheringState.COMPLETE) return@collect
+                    wrapper.getIceCredentials().onEach { (mid, credentials) ->
```
tpettersen, line 286: Should all candidates send empty sdp?

tpettersen (approved):
Lgtm! Small question just to understand

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
drymarau: There should be one `end-of-candidates` (in this case an ICE candidate with an empty SDP) for each transceiver.

PR #223, opened 2023-10-10
feat(media): SecureCheckCode

drymarau:


vadymmarkov (approved):
Looks good. I'll also adjust the algorithm in Swift SDK so we produce the same values.

PR #225, opened 2023-10-11
feat: bump gradle to 8.4

drymarau:


tpettersen (approved):
Builds and runs here 👍

PR #226, opened 2023-10-12
feat(api): add several direct media events

drymarau:
Skipping `splash_screen` for the time being.


sdk-api-infinity/src/main/kotlin/com/pexip/sdk/api/infinity/Event.kt
```diff
@@ -22,6 +22,18 @@ import kotlinx.serialization.Serializable
...
 import java.util.UUID
 
+@Serializable
+public data class NewOfferEvent(val sdp: String) : Event
+
+@Serializable
+public data class UpdateSdpEvent(val sdp: String) : Event
+
+@Serializable
+public data class NewCandidateEvent(val candidate: String, val mid: String) : Event
```
tpettersen, line 32: Why not deserialize `ufrag` and `pwd` also? You dont need it?

tpettersen (approved):
Looks good 👍 Small question only

sdk-api-infinity/src/main/kotlin/com/pexip/sdk/api/infinity/Event.kt
drymarau: `ufrag` is already embedded in the `candidate` and `pwd` is not sent over SSE. I believe the need for both is due to catering to JS API which exposes both, but it's not required for Android.

We can add `ufrag` through, if only to reduce the number of breaking changes later if we ever decide to use it.

PR #227, opened 2023-10-12
feat(api): add AckRequest

drymarau:


sdk-api-infinity/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
```diff
@@ -552,6 +552,32 @@ public interface InfinityService {
          */
         public fun ack(token: Token): Call<Unit>
 
+        /**
+         * Acks the call.
```
tpettersen, line 556: You correctly point to the docs, but it would not hurt to mention here that the AckRequest is for direct media and it should contain the local SDP - only to avoid confusion :)

sdk-api-infinity/src/main/kotlin/com/pexip/sdk/api/infinity/internal/CallStepImpl.kt
```diff
@@ -80,6 +81,28 @@ internal class CallStepImpl(
 
     override fun ack(token: Token): Call<Unit> = ack(token.token)
 
+    override fun ack(request: AckRequest, token: String): Call<Unit> {
+        require(token.isNotBlank()) { "token is blank." }
+        return RealCall(
+            client = client.newBuilder()
+                .readTimeout(0, TimeUnit.SECONDS)
```
tpettersen, line 88: Any reason for not having a timeout?

tpettersen (approved):
Seems to be at spec 👍

sdk-api-infinity/src/main/kotlin/com/pexip/sdk/api/infinity/internal/CallStepImpl.kt
drymarau: https://github.com/pexip/pexip-android-sdk/pull/135#issue-1792980589

PR #236, opened 2023-10-20
feat(api,conference): add directMediaRequested property

drymarau:
This is a trick to avoid callers of `Referer.refer()` of passing `directMedia`. Instead it uses the value that the original request had.


tpettersen (approved):
There is a test improvement that would be good to fix. Other than that it seems like a smart approach to do the tagging to pass this prop 👌

PR #239, opened 2023-11-03
feat(media): add onFirstFrame and onAspectRatioChange

drymarau:
This should help build UIs that react to changes in aspect ratio

### DIFF
--- a/sample/src/main/kotlin/com/pexip/sdk/sample/conference/ConferenceCallScreen.kt
+++ b/sample/src/main/kotlin/com/pexip/sdk/sample/conference/ConferenceCallScreen.kt
@@ -50,6 +50,7 @@ import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableFloatStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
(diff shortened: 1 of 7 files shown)


sample/src/main/kotlin/com/pexip/sdk/sample/conference/ConferenceCallScreen.kt
```diff
@@ -139,10 +142,12 @@ fun ConferenceCallScreen(
                     modifier = Modifier
                         .align(Alignment.TopEnd)
                         .fillMaxWidth(0.25f)
-                        .aspectRatio(aspectRatio),
+                        .then(aspectRatioModifier),
```
tpettersen, line 145: For selfView I am seeing som odd things on tablet emulator (pixel C - API 33) - which I cannot reproduce on my phone. Are you able to repro?

When doing orientation changes, I get wrong aspect ratios for selfView in portrait (and even wrong orientatition in normal portrait - but correct in reverse portrait) and upside-down in one landscape: 
image
image

image

drymarau: Seems like a problem in the emulator, this PR doesn't rotate the frames (which is pretty evident at least in some of these screenshots).

tpettersen: But it does do aspect ratio, which is failing in the emulator for some reason. 🤔 
I will do a bit more testing then

drymarau: It just means that emulator is capturing in landscape when it's in portrait.

tpettersen (approved):
Works on my phone :)

PR #241, opened 2023-11-03
refactor(conference): move coroutines extensions to Messenger

drymarau:


sdk-conference/src/main/kotlin/com/pexip/sdk/conference/coroutines/Util.kt
```diff
@@ -51,31 +46,20 @@ public fun Conference.getConferenceEvents(): Flow<ConferenceEvent> = callbackFlo
  * @return a message
  * @throws MessageNotSentException when there was an issue when sending the message
  */
+@Deprecated(message = "Use the member method instead.", level = DeprecationLevel.ERROR)
```
tpettersen, line 49: Jumping straight to error? Might aswell just add a `throw` instead of `send(type, payload, participantId)` also then

tpettersen (approved):
Works good 👍 Minor issue with paddings which you can address if you want

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/coroutines/Util.kt
drymarau: It can't be called anyway

PR #245, opened 2023-11-14
feat(media): direct media

drymarau:
This PR implements direct media support for the SDK.

It includes perfect negotiation, as well as some changes necessary for seamless handling of peer disconnection.

Note that the theme API will come separate PR.

### DIFF
--- a/sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/AbstractMessenger.kt
+++ b/sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/AbstractMessenger.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2023 Pexip AS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.pexip.sdk.conference.infinity.internal
+
+import com.pexip.sdk.conference.Message
+import com.pexip.sdk.conference.MessageListener
+import com.pexip.sdk.conference.MessageNotSentException
+import com.pexip.sdk.conference.Messenger
+import com.pexip.sdk.conference.SendCallback
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import java.util.UUID
+import java.util.concurrent.CopyOnWriteArraySet
+
+internal abstract class AbstractMessenger(private val scope: CoroutineScope) : Messenger {
+
+    private val listeners = CopyOnWriteArraySet<MessageListener>()
+
+    @Deprecated("Use the coroutines version of this method.")
+    override fun send(type: String, payload: String, callback: SendCallback) =
+        sendInternal(type, payload, callback)
+
+    @Deprecated("Use the coroutines version of this method.")
+    override fun send(participantId: UUID, type: String, payload: String, callback: SendCallback) =
+        sendInternal(type, payload, callback, participantId)
+
+    @Deprecated("Use getMessages() that returns a Flow")
+    override fun registerMessageListener(listener: MessageListener) {
+        listeners += listener
+    }
+
+    @Deprecated("Use getMessages() that returns a Flow")
+    override fun unregisterMessageListener(listener: MessageListener) {
+        listeners -= listener
+    }
+
+    protected fun onMessage(message: Message) = listeners.forEach { it.onMessage(message) }
+
+    private fun sendInternal(
+        type: String,
+        payload: String,
+        callback: SendCallback,
+        participantId: UUID? = null,
+    ) {
+        scope.launch {
+            try {
+                val message = send(type, payload, participantId)
+                callback.onSuccess(message)
+            } catch (e: MessageNotSentException) {
+                callback.onFailure(e)
+            }
+        }
+    }
+}
--- a/sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/DataChannelImpl.kt
+++ b/sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/DataChannelImpl.kt
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2023 Pexip AS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.pexip.sdk.conference.infinity.internal
+
+import com.pexip.sdk.media.Data
+import com.pexip.sdk.media.DataChannel
+import com.pexip.sdk.media.DataSender
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.asSharedFlow
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
+internal class DataChannelImpl(override val id: Int) : DataChannel {
+
+    private val mutex = Mutex()
+    private val _data = MutableSharedFlow<Data>()
+
+    private var sender: DataSender? = null
+
+    override val data: Flow<Data> = _data.asSharedFlow()
+
+    override suspend fun send(data: Data) = mutex.withLock {
+        val sender = checkNotNull(sender) { "sender is not attached." }
+        sender.send(data)
+    }
+
+    suspend fun attach(sender: DataSender) = mutex.withLock {
+        this.sender = sender
+    }
+
+    suspend fun detach(sender: DataSender) = mutex.withLock {
+        if (this.sender == sender) {
+            this.sender = null
+        }
+    }
+
+    suspend fun onData(data: Data) = _data.emit(data)
+}
--- a/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/Observer.kt
+++ b/sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/Observer.kt
@@ -15,6 +15,7 @@
  */
 package com.pexip.sdk.media.webrtc.internal
 
+import com.pexip.sdk.media.Data
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.asSharedFlow
 import org.webrtc.CandidatePairChangeEvent
@@ -54,7 +55,11 @@ internal class Observer(private val continualGatheringPolicy: ContinualGathering
     }
 
     override fun onMessage(buffer: DataChannel.Buffer) {
-        _event.tryEmit(Event.OnMessage(buffer.data))
+        val data = Data(
+            data = ByteArray(buffer.data.remaining(), buffer.data::get),
+            binary = buffer.binary,
+        )
+        _event.tryEmit(Event.OnData(data))
     }
 
     override fun onRenegotiationNeeded() {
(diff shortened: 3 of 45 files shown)


sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
```diff
@@ -282,11 +323,58 @@ internal class WebRtcMediaConnection(
...
+                    val candidate = IceCandidate(event.mid, -1, event.candidate)
+                    wrapper.addIceCandidate(candidate)
+                } catch (e: CancellationException) {
+                    throw e
+                } catch (t: Throwable) {
+                    if (mutex.withLock { !ignoreOffer }) {
+                        throw t
+                    }
+                }
+                is RestartSignalingEvent -> {
```
vadymmarkov, line 371: What is this `RestartSignalingEvent`? I don't think I'm doing anything similar in the Swift SDK.

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
```diff
@@ -370,25 +458,30 @@ internal class WebRtcMediaConnection(
...
 
         override val mediaType: MediaType = MediaType.MEDIA_TYPE_AUDIO
+        override val streamIds: List<String> = listOf("main")
     }
 
     private data object MainVideo : RtpTransceiverKey {
 
         override val mediaType: MediaType = MediaType.MEDIA_TYPE_VIDEO
+        override val streamIds: List<String> = listOf("main")
+        override val sendEncodings: List<RtpParameters.Encoding> =
```
vadymmarkov, line 476: Why do you need to keep a list of streamIds and encodings here? In the Swift SDK we sync transceivers by setting a sender track, but not this.

sdk-media/src/main/kotlin/com/pexip/sdk/media/MediaConnectionSignaling.kt
```diff
@@ -15,31 +15,56 @@
...
+ * @property iceTransportsRelayOnly whether relay only mode should be used
+ * @property event a [Flow] of [SignalingEvent]s
+ * @property dataChannel an optional [DataChannel] for messaging between peers
  */
 public interface MediaConnectionSignaling {
 
-    /**
-     * A list of available [IceServer]s.
-     */
+    public val directMedia: Boolean
```
vadymmarkov, line 31: Any particular reason `directMedia` and `iceTransportsRelayOnly` are here and not in `MediaConnectionConfig`?

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/PeerConnectionWrapper.kt
```diff
@@ -119,40 +192,95 @@ internal class PeerConnectionWrapper(factory: PeerConnectionFactory, rtcConfig:
...
+    // for native-managed pointers.
+    private inline fun PeerConnection.withTransceivers(block: (RtpTransceiver) -> Unit) {
+        val oldTransceivers = currentTransceivers
+        val newTransceivers = nativeTransceivers
+        currentTransceivers = newTransceivers
+        newTransceivers.forEach(block)
+        oldTransceivers.forEach {
+            val dtmfSender = it.sender.dtmf()
+            if (dtmfSender != null) {
+                safeNativeReleaseRef(dtmfSender.ref)
```
vadymmarkov, line 264: Is it like manual reference counting?

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/PeerConnectionWrapper.kt
```diff
@@ -24,54 +25,126 @@ import kotlinx.coroutines.flow.filterIsInstance
...
+            null -> null
+            else -> peerConnection!!.createDataChannel("pexChannel", init)
+        }
+    }
+
+    suspend fun start() = mutex.withLock {
+        observer.start()
+    }
+
+    suspend fun restart() = mutex.withLock {
```
vadymmarkov, line 95: Do you use this `restart` function instead of call transfer?

sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/AbstractMessenger.kt
```diff
@@ -0,0 +1,67 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import java.util.UUID
+import java.util.concurrent.CopyOnWriteArraySet
+
+internal abstract class AbstractMessenger(private val scope: CoroutineScope) : Messenger {
+
+    private val listeners = CopyOnWriteArraySet<MessageListener>()
+
+    @Deprecated("Use the coroutines version of this method.")
```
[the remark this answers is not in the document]
drymarau: It's actually deprecated in the `Messenger` interface and deprecations should usually be inherited.

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/Util.kt
```diff
@@ -157,4 +161,6 @@ internal fun Encoding(
 
 internal const val MAX_FRAMERATE = 30
 
+internal val ValidDataChannelIds = 0..65535
```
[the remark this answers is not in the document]
drymarau: Good catch, I don't believe it's used any longer.

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
drymarau: This is a different name for `peer_disconnect` event, which I believe is a bit more descriptive of what the consumer should do - restart the underlying `PeerConnection`.

sdk-media/src/main/kotlin/com/pexip/sdk/media/MediaConnectionSignaling.kt
drymarau: Both are derived from the conference token response and happen to be read-only. It might be useful to add an option to override `iceTransportsRelayOnly` in `MediaConnectionConfig` for transcoded calls, although I don't think we have a use-case for this.

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/PeerConnectionWrapper.kt
drymarau: Yeah, but even more dirty. There's a short comment explaining why it has to be this way over the method signature.

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/PeerConnectionWrapper.kt
drymarau: Yes, it ended up being feasible to just create a new underlying `PeerConnection` instead of a full-blown call transfer since we only need to ensure that our DTLS keys are not re-used. It should help consumers with implementing this feature as there's a very thin API surface (it should be enough to just put `directMedia = true` in all `RequestTokenRequest` calls for direct media to work).

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/Observer.kt
```diff
@@ -28,19 +29,43 @@ import org.webrtc.RtpReceiver
...
+    private val started = AtomicBoolean()
     private val candidates = mutableMapOf<String, IceCandidate>()
-    private val renegotiationNeeded = AtomicBoolean(false)
     private val _event = MutableSharedFlow<Event>(extraBufferCapacity = 64)
 
     val event = _event.asSharedFlow()
 
+    fun start() {
+        if (started.compareAndSet(false, true)) {
+            _event.tryEmit(Event.OnRenegotiationNeeded)
```
tpettersen, line 43: I dont quite understand this. Why do we need renegotiation on the first start? Has it been negotiated already?

sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/DataChannelImpl.kt
```diff
@@ -0,0 +1,52 @@
...
+    private val _data = MutableSharedFlow<Data>()
+
+    private var sender: DataSender? = null
+
+    override val data: Flow<Data> = _data.asSharedFlow()
+
+    override suspend fun send(data: Data) = mutex.withLock {
+        val sender = checkNotNull(sender) { "sender is not attached." }
+        sender.send(data)
+    }
```
tpettersen, line 39: Would have been nifty if the sender itself was available to the user/code so it could just send directly through it instead of through the data channel. Then you would not have the potential case where sender is not attached. 
Not important at all though :)

tpettersen (approved):
I left some comments and suggestions, but overall this seems to work solid 👏 Great work

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/PeerConnectionWrapper.kt
```diff
@@ -173,8 +313,98 @@ internal class PeerConnectionWrapper(factory: PeerConnectionFactory, rtcConfig:
...
+        get() = NativeRtpTransceiver.getLong(this)
+
+    private fun RtpReceiver.nativeUnsetObserver(ref: Long, observerRef: Long) =
+        NativeUnsetObserver.invoke(this, ref, observerRef)
+
     companion object {
 
+        private val NativeGetTransceivers by lazy {
+            PeerConnection::class.java
+                .getDeclaredMethod("nativeGetTransceivers")
```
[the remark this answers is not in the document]
drymarau: They're going to have a really bad time. But that's also true if you replace the version on `main` with a different one - even public methods may come and go.

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/PeerConnectionWrapper.kt
```diff
@@ -173,8 +313,98 @@ internal class PeerConnectionWrapper(factory: PeerConnectionFactory, rtcConfig:
...
+            RtpReceiver::class.java
+                .getDeclaredField("nativeRtpReceiver")
+                .also { it.isAccessible = true }
+        }
+        private val NativeRtpTransceiver by lazy {
+            RtpTransceiver::class.java
+                .getDeclaredField("nativeRtpTransceiver")
+                .also { it.isAccessible = true }
+        }
+
```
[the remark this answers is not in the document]
drymarau: Great idea, should be fairly easy to do.

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/Observer.kt
drymarau: It's just a single to kickstart the whole process of negotiation; when `PeerConnection` is first created, the `onRenegotiationNeeded()` will fire, but we want to skip that one and let `start()` handle it instead.

sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/DataChannelImpl.kt
drymarau: That's just the nature of things when we have to move messages through the `DataChannel` instead of `Conference` - there may simply be no `DataChannel` available.

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/PeerConnectionWrapper.kt
```diff
@@ -24,54 +25,126 @@ import kotlinx.coroutines.flow.filterIsInstance
...
 import org.webrtc.VideoTrack
+import java.nio.ByteBuffer
 import kotlin.coroutines.Continuation
 import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
+import kotlin.properties.Delegates
 
-internal class PeerConnectionWrapper(factory: PeerConnectionFactory, rtcConfig: RTCConfiguration) {
+internal class PeerConnectionWrapper(
```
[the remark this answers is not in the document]
drymarau: Might require a bit more investigation, presumably that black frame we insert should've been enough.

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/WebRtcMediaConnection.kt
```diff
@@ -100,8 +111,18 @@ internal class WebRtcMediaConnection(
 
     init {
         with(scope) {
+            launchDataSender()
+            if (config.signaling.directMedia) {
```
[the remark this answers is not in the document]
drymarau: This ensures a few things:
- Stable order of transceivers
- Data channel having the correct placement (if it's any other `mid` it fails to establish connection with the other side, even with correctly specified `id`)

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/internal/PeerConnectionWrapper.kt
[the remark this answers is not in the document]
drymarau: Ideally this should be handled based on roster list and the other participant state and the `SurfaceView` should go away to reveal the splash screen (there's indeed one for when the video is not available).

PR #246, opened 2023-11-16
feat(api,conference): add splash screens for direct media calls

drymarau:


sdk-api-infinity/src/main/kotlin/com/pexip/sdk/api/infinity/internal/ConferenceStepImpl.kt
```diff
@@ -128,6 +129,37 @@ internal class ConferenceStepImpl(
...
+                .build(),
+            mapper = ::parseTheme,
+        )
+    }
+
+    override fun theme(token: Token): Call<Map<String, SplashScreenResponse>> = theme(token.token)
+
+    override fun theme(path: String, token: String): String {
+        require(path.isNotBlank()) { "path is blank." }
+        return node.newApiClientV2Builder()
```
vadymmarkov, line 153: Do you have multiple builders?

vadymmarkov (approved):
Looks good. Tested and the splash screen shows up in the sample app.

sdk-conference-infinity/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImpl.kt
```diff
@@ -0,0 +1,79 @@
...
+                elements = response.elements.mapNotNull {
+                    when (it) {
+                        is ElementResponse.Text -> Element.Text(it.color, it.text)
+                        is ElementResponse.Unknown -> null
+                    }
+                },
+                background = step.theme(response.background.path, store.get()),
+            )
+        }
+    }
```
tpettersen, line 71:
```suggestion
    ) = event.screenKey?.let { key ->
        val response = responses[key] ?: return null
        SplashScreen(
            key = key,
            elements = response.elements.mapNotNull {
                when (it) {
                    is ElementResponse.Text -> Element.Text(it.color, it.text)
                    is ElementResponse.Unknown -> null
                }
            },
            background = step.theme(response.background.path, store.get()),
        )
    }
```

Nitpick: a bit hard to read with nested whens and nulls

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/SplashScreen.kt
```diff
@@ -0,0 +1,31 @@
...
+ * Only available in direct media calls.
+ *
+ * @property key a unique key that identifies this [SplashScreen]
+ * @property elements a list of [Element]s that should be drawn over the [background]
+ * @property background a URL that contains the background image
+ */
+public data class SplashScreen(
+    val key: String,
+    val elements: List<Element>,
+    val background: String,
```
tpettersen, line 30: nitnitnitpick: does not hurt to have `backgroundUrl` or `backgroundPath` as the variable name or even type `URL` to underline that it is indeed a URL we are dealing with.

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Element.kt
```diff
@@ -0,0 +1,30 @@
...
+ */
+public sealed interface Element {
+
+    /**
+     * A text element.
+     *
+     * @property color the color of this [Text]
+     * @property text the text to display
+     */
+    public data class Text(val color: Long, val text: String) : Element
```
tpettersen, line 29: Missing quite a few properties for Text which is webby, but at least font_size and line_spacing might be relevant for the clients.
```
{
  "type": "text",
  "label": "enter_conference_id",
  "offset_y": 50,
  "offset_x": 0,
  "height": 200,
  "width": 1920,
  "color": "0xffffffff",
  "font_size": 66,
  "font": "GT Pressura Light",
  "letter_spacing": 0,
  "line_spacing": -13,
  "align": "center"
}
```

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Element.kt
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.pexip.sdk.conference
+
+/**
+ * A drawable element of a [SplashScreen].
+ */
+public sealed interface Element {
```
tpettersen, line 21: What about `icon` and `pin_digits`? No reason to support them perhaps?

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Element.kt
drymarau: I don't believe these are used for `direct_media_` screens, at least the events themselves do not contain these properties.

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/SplashScreen.kt
drymarau: Most image-loading libraries should accept `String` just fine. Unless you want to save the splash screen, of course. I'd like to migrate to `HttpUrl`, but I'm not entirely sure whether to do it now or later.

sdk-api-infinity/src/main/kotlin/com/pexip/sdk/api/infinity/internal/ConferenceStepImpl.kt
drymarau: Just this one, but now used in several places.

tpettersen (approved):

tpettersen:
Small bug report:
1. Start call and the splash shows
2. Join with 2 participants and get a multi-party-conference
3. leave the call with those same 2. 

The splash shows for a couple of seconds, and then disappears - leaving a black screen

drymarau:
> Small bug report:
> 
> 1. Start call and the splash shows
> 2. Join with 2 participants and get a multi-party-conference
> 3. leave the call with those same 2.
> 
> The splash shows for a couple of seconds, and then disappears - leaving a black screen

I'm unable to reliably reproduce it - did it only once out of several tries. I wonder if eagerly subscribing is the way to go here in case we lose the event?

drymarau:
> Feels like a race IMO 🤔 Or did you find the root cause?

Just an assumption for now, set the `Theme.spashScreen` to `SharidStarted.Eagerly` and we'll see how it goes.

PR #251, opened 2023-11-22
refactor: merge base modules with -infinity modules

drymarau:
In a similiar fashion to what we previously did with `-coroutines` modules, we now merge `-infinity` modules with their respective bases. This should allow us to re-use most classes across higher level constructs, as well as to converge on a single module that deals with Infinity-related code - `:sdk-infinity`, which has now becomes the base.


vadymmarkov (approved):
Does it mean there is only one `sdk-infinity` module now? Any plans on merging that with `sdk-conference` as well?

drymarau:
> Does it mean there is only one `sdk-infinity` module now? Any plans on merging that with `sdk-conference` as well?

No modules were removed in this PR. But the overall direction would be to remove `sdk-api`/`sdk-conference`/`sdk-registration` (and all `-coroutines` & `-infinity` modules). Right now `sdk-infinity` is a base of `sdk-api`, on which both `sdk-conference` and `sdk-registration` depend on.

tpettersen (approved):

PR #254, opened 2023-11-24
refactor(api,conference): move data channel message serialization to :sdk-api

drymarau:


sdk-conference/src/test/kotlin/com/pexip/sdk/conference/infinity/internal/DataChannelMessengerImplTest.kt
```diff
@@ -189,14 +171,15 @@ class DataChannelMessengerImplTest {
         messenger.message.test {
             val messages = List(10) { Random.nextMessage(at) }
             messages.forEach {
-                val body = MessageBody(
-                    type = it.type,
-                    payload = it.payload,
-                    uuid = it.participantId,
-                    origin = it.participantName,
+                val message = DataChannelMessage.Message(
```
tpettersen, line 174: Would be good to also have a test case for `DataChannelMessage.Unknown` since that has a throw

tpettersen (approved):
Works good!

PR #258, opened 2023-11-27
refactor(conference,registration): throw on unsupported Infinity version

drymarau:


sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/InfinityConference.kt
```diff
@@ -183,14 +182,13 @@ public class InfinityConference private constructor(
             response: RequestTokenResponse,
         ): InfinityConference {
             if (response.version.versionId < "29") {
-                val logger = Logger.getLogger("InfinityConference")
-                val msg = buildString {
+                val s = buildString {
```
tpettersen, line 185: It works, but IMO it would make more sense to throw down at the `parseRequestToken` in the API call to stop users in their tracks as fast as possible. Now it is possible to create a token and be "alive" until you create a conference with it.

tpettersen (approved):
I would have preferred to move the throw as low as possible, but I will leave that up to you. 👍 

Couple of things I was thinking about:
- is it possible to refer to another installation? And if so, how will the logic behave if that installation is obsolete?
- It would be a strictly nice-to-have to show the error in PIN challenge / requirement as a popup in the sample app, instead of a "incorrect PIN" thingy.

drymarau:
> * It would be a strictly nice-to-have to show the error in PIN challenge / requirement as a popup in the sample app, instead of a "incorrect PIN" thingy.

I think it already shows a toast on all other errors

PR #260, opened 2023-11-28
refactor(media): replace onAspectRatioChange with onFrameResolutionChange

drymarau:


sdk-media-webrtc-compose/src/main/kotlin/com/pexip/sdk/media/webrtc/compose/VideoTrackRenderer.kt
```diff
@@ -54,7 +54,7 @@ public fun VideoTrackRenderer(
     zOrderMediaOverlay: Boolean = false,
     zOrderOnTop: Boolean = false,
     onFirstFrame: () -> Unit = { },
-    onAspectRatioChange: (Float) -> Unit = { },
+    onFrameResolutionChange: (width: Int, height: Int, rotation: Int) -> Unit = { _, _, _ -> },
```
tpettersen, line 57: would not hurt to include aspectRatio as a fourth param here, to avoid clients having to calculate every time based on the rotation

tpettersen (approved):

PR #264, opened 2023-11-29
feat(api): bump readTimeout to 1 minute

drymarau:
10 seconds aren't always enough in some circumstances

### DIFF
--- a/sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
+++ b/sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
@@ -723,13 +723,15 @@ public interface InfinityService {
         @JvmStatic
         @JvmOverloads
         public fun create(client: OkHttpClient = OkHttpClient()): InfinityService = create(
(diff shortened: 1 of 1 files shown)


sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
```diff
@@ -722,7 +723,9 @@ public interface InfinityService {
         @JvmStatic
         @JvmOverloads
         public fun create(client: OkHttpClient = OkHttpClient()): InfinityService = create(
-            client = client,
+            client = client.newBuilder()
+                .readTimeout(1, TimeUnit.MINUTES)
```
tpettersen, line 727: Would it make sense to make this configurable? I do not know if there is some client security demands of shorter read timeouts or similar

tpettersen (approved):
👌 could also consider setting `callTimeout`

drymarau:
> 👌 could also consider setting `callTimeout`

That's unnecessary, it's already set to 0.

PR #275, opened 2023-12-06
feat: bump webrtc to 119.0.0

drymarau:


tpettersen (approved):
Works good when testing video

PR #276, opened 2023-12-07
refactor: switch back to kapt

drymarau:
Some issues with Dagger not picking up Wire-generated classes after update to Kotlin 1.9.2x


tpettersen (approved):
Looking forward for a stable dagger ksp support

PR #277, opened 2023-12-07
refactor(media): move mode and focus handling to ADM

drymarau:


sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/WebRtcMediaConnectionFactory.kt
```diff
@@ -339,8 +347,44 @@ public class WebRtcMediaConnectionFactory private constructor(
...
+                            // noop
+                        }
+                        .build()
+
+                override fun onWebRtcAudioTrackStart() {
+                    AudioManagerCompat.requestAudioFocus(audioManager, request)
+                }
+
+                override fun onWebRtcAudioTrackStop() {
+                    AudioManagerCompat.abandonAudioFocusRequest(audioManager, request)
```
tpettersen, line 366: This is not working for me. I am playing spotify, enters a call and loses the music - all expected. But if I then leave the conference the music does not start playing again. That happens on `main`

tpettersen (changes requested):

sdk-media-webrtc/src/main/kotlin/com/pexip/sdk/media/webrtc/WebRtcMediaConnectionFactory.kt
drymarau: There was a change in what type of gain we use - used to be transient, now regular. Transient will resume the playback, but the regular one signifies permanent loss of focus.

I'd argue that for a short call transient is better, but for an hour long meeting your previous resuming playback might be both unexpected and undesirable.

tpettersen: True 🤔

tpettersen (approved):

PR #291, opened 2023-12-20
refactor(sample): extract chat into a separate workflow

drymarau:


tpettersen (changes requested):
It looks really good, but there is a bug that should be addressed. When a third user enters the call, you get transferred from direct media to a multi-party conference. Then the `ChatProps` changes to use a `MessengerImpl` instead of the `DataChannelMessengerImpl`. When that happens, the client is no longer able to send messages, as the send-button never gets clickable.

drymarau:
Good catch, should be fixed now I believe.

tpettersen (approved):
Works good now

PR #310, opened 2024-02-07
fix(api): do not throw from ParticipantResponseSerializer, return JsonNull instead

drymarau:
`coerceInputValues` property does not seem to be honored when the serialiazer throws `SerializationException`.

We wil now return `JsonNull` instead, which should fallback to a default value defined in the `ParticipantResponse` itself.


tpettersen (approved):
It works when testing with lauda, so I am happy!

PR #318, opened 2024-02-13
refactor(media): do not require BLUETOOTH_CONNECT permission

drymarau:
`AudioManager.setCommunicationDevice()` seems to be handling most of the work and does not require this permission to connect and use a Bluetooth headset.


tpettersen (approved):
Works alright 👍

PR #322, opened 2024-02-15
feat(conference): add Theme.layout

drymarau:
This should provide read-only access to currently active layout of the conference.

Ability to transform the layout will come at a later stage.

### DIFF
--- a/sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/Event.kt
+++ b/sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/Event.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022-2023 Pexip AS
+ * Copyright 2022-2024 Pexip AS
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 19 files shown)


sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/Screen.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+public data class Screen(
+    @SerialName("chair_layout")
+    val hostLayout: LayoutId,
+    @SerialName("guest_layout")
+    val guestLayout: LayoutId,
```
tpettersen, line 26: According to the docs it is supposed to be "chair" and "guest" 🤔 Based on the response you use in the tests, I guess the docs are misleading

sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/Event.kt
```diff
@@ -80,6 +80,16 @@ public data class MessageReceivedEvent(
     val direct: Boolean = false,
 ) : Event
 
+@Serializable
+public data class LayoutEvent(
```
tpettersen, line 84: No need for participants yet?

tpettersen (approved):
Looks good :)

sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/Event.kt
drymarau: No clear idea how to use that property in an Android client to be honest. Easy to add later with an empty `Set` as a default value without breaking the clients.

sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/Screen.kt
drymarau: Test file is from Infinity, while I'm not sure where the one in docs comes from.

PR #323, opened 2024-02-15
feat(media): ICE restart on connection type change

drymarau:
This should resilience when switching between different networks. Continual gathering occassionaly ended missing audio during my testing when switching.

Note that in some cases it may take a few seconds to find new ICE candidates and establish the connection.


tpettersen (changes requested):
This is not working 100% for me when testing. Sending of camera resumes, but receiving video does not. It is stuck on last frame.

drymarau:
> This is not working 100% for me when testing. Sending of camera resumes, but receiving video does not. It is stuck on last frame.

Need a bit more info on the setup. Also, how long have you waited for the video to resume?

drymarau:
Can you use a real Android device? Not entirely whether this behavior can be attributed to the emulator or the change in this PR.

tpettersen:
It works way better on a real device 👌 Only thing I see there is that it occasionally crashes with this:
```
15:14:28.590  D  --> POST https://nightly.pexip.com/api/client/v2/conferences/meet.thomas.pettersen@nightly.pexip.com/participants/69f5331c-dd77-498f-b92b-fe8f85b01566/calls/ddb0b28c-7518-452d-87d3-04d8bb0eb7ed/update (3587-byte body)
15:14:28.590  I  0xb4000074ed137380 c2.qti.avc.encoder 704x1280 inputFps=29 outputFps=30 renderFps=0, discardFps=30
15:14:28.592  D  <-- HTTP FAILED: java.net.UnknownHostException: Unable to resolve host "nightly.pexip.com": No address associated with hostname
15:14:28.620  E  FATAL EXCEPTION: WebRtcMediaConnectionFactory
                 Process: com.pexip.sdk.sample, PID: 31603
                 
```

I reckon this is due to a minor loss of actual network coverage when switching the moment a POST call is made.

drymarau:
> I reckon this is due to a minor loss of actual network coverage when switching the moment a POST call is made.

DNS might also be to blame. I believe this might be reproducible on `main` as well.

tpettersen (approved):

PR #329, opened 2024-02-20
feat(conference): expose layout SVGs on Layout

drymarau:
Note that this will be a raw `String` and an appropriate library must be used to render it.

### DIFF
--- a/sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
+++ b/sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
@@ -180,6 +180,28 @@ public interface InfinityService {
         public fun availableLayouts(token: String): Call<Set<LayoutId>>
 
         /**
+         * This provides all SVG representations of the layouts that are active
+         * on the given conference.
+         *
+         * See [documentation](https://docs.pexip.com/beta/api_client/api_rest.htm#layout_svgs).
+         *
+         * @param token a valid token
+         * @return a collection of SVG representations
+         */
+        public fun layoutSvgs(token: Token): Call<Map<LayoutId, String>>
+
+        /**
+         * This provides all SVG representations of the layouts that are active
+         * on the given conference.
+         *
+         * See [documentation](https://docs.pexip.com/beta/api_client/api_rest.htm#layout_svgs).
+         *
+         * @param token a valid token
+         * @return a colleciton of SVG representations
+         */
+        public fun layoutSvgs(token: String): Call<Map<LayoutId, String>>
+
+        /**
          * This returns a list of all available layouts for the given conference.
          *
          * This includes the inbuilt layouts plus any custom layouts available on this conference.
--- a/sdk-api/src/test/kotlin/com/pexip/sdk/api/infinity/ConferenceStepTest.kt
+++ b/sdk-api/src/test/kotlin/com/pexip/sdk/api/infinity/ConferenceStepTest.kt
@@ -489,6 +489,69 @@ internal class ConferenceStepTest {
     }
 
     @Test
+    fun `layoutSvgs throws IllegalStateException`() = runTest {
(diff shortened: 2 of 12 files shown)


sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
```diff
@@ -179,6 +179,28 @@ public interface InfinityService {
          */
         public fun availableLayouts(token: String): Call<Set<LayoutId>>
 
+        /**
+         * This provides all SVG representations of the layouts that are active
+         * on the given conference.
+         *
+         * See [documentation](https://docs.pexip.com/beta/api_client/api_rest.htm#layout_svgs).
```
tpettersen, line 186: I am a bit confused. Will the layout_svgs be available on all nodes? Is there a minimum level? And why are we referencing beta docs?

drymarau: No, it will only be available starting with v34. The implementation in `Theme` will return an empty `Map` on 404.

tpettersen: On that note, maybe it would be an idea to set "supported from Infinity version 34" or similar on all APIs, to note what version is supported. Akin to kdocs for android

tpettersen (approved):

sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
drymarau: It might be a bit hard to enforce it without any tooling, especially given that it's a runtime check.

tpettersen: No enforcing, just a courtesy for the developers through docs

drymarau: I think we can rely on the links to main website since it does include information when the API has been added.

I'll see if Infinity gives something useful when the method is missing, so maybe we could throw a different `Exception` in such a case.

drymarau: It serves an HTML page when a resource does not exist:
```html
<head>
    <title>404 - No Such Resource</title>
</head>                                                                                                    
<body>                                                                                                    
    <h1>No Such Resource</h1>                                                                                                    
    <p>No such child resource.</p>                                                                                                  
</body>                                                                                          
```

I'm not sure how to approach it, but the naive solution of checking whether the response body starts with `<html>` might be OK (since we already match on code).

sdk-api/src/test/kotlin/com/pexip/sdk/api/infinity/ConferenceStepTest.kt
```diff
@@ -631,6 +694,17 @@ internal class ConferenceStepTest {
         assertGet()
     }
 
+    private fun MockWebServer.verifyLayoutSvgs(token: String) = takeRequest {
+        assertRequestUrl(node) {
+            addPathSegments("api/client/v2")
+            addPathSegment("conferences")
+            addPathSegment(conferenceAlias)
+            addPathSegment("layout_svgs")
```
vadymmarkov, line 702: Is this endpoint documented somewhere?

drymarau: There's a link to beta docs in this method's KDoc.

drymarau: It only appeared in v34, which is yet to be released.

PR #330, opened 2024-02-20
feat(api): add layout_transform

drymarau:
This is a very first revision with only few available properties.


drymarau:
One can test it by adding the following few lines in `InfinityConference.init`:
```kotlin
scope.launch {
    while (isActive) {
        delay(5.seconds)
        val layouts = step.availableLayouts(store.get()).await()
        val request = TransformLayoutRequest(
            hostLayout = layouts.random(),
            enableOverlayText = Random.nextBoolean(),
        )
        step.transformLayout(request, store.get()).await()
    }
}
```

tpettersen (approved):
Works good :) Even got the crash when setting guest_layout

PR #331, opened 2024-02-24
fix(conference): certain components missing SSE

drymarau:
This was most obvious with the recent addition of `Layout`. The SSE would arrive before the `Theme.layout` had a chance to subscribe and the current "state" would end up being ignored.

This PR forces the underlying `event` to wait until all expected subscribers are active to initiate the connection.

An alternative implementation that relied on debouncing was considered, but was dismissed due to being unreliable (hard to guess the correct timeout value).

### DIFF
--- a/sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/InfinityConference.kt
+++ b/sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/InfinityConference.kt
@@ -35,6 +35,7 @@ import com.pexip.sdk.conference.infinity.internal.MessengerImpl
 import com.pexip.sdk.conference.infinity.internal.RefererImpl
 import com.pexip.sdk.conference.infinity.internal.RosterImpl
 import com.pexip.sdk.conference.infinity.internal.ThemeImpl
+import com.pexip.sdk.conference.infinity.internal.WhileSubscribedAtLeast
 import com.pexip.sdk.conference.infinity.internal.events
 import com.pexip.sdk.infinity.UnsupportedInfinityException
 import com.pexip.sdk.media.IceServer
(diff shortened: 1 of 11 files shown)


sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/InfinityConference.kt
```diff
@@ -64,7 +65,12 @@ public class InfinityConference private constructor(
     private val executor = Executors.newSingleThreadScheduledExecutor()
     private val scope = CoroutineScope(SupervisorJob() + executor.asCoroutineDispatcher())
     private val store = TokenStore.create(response)
-    private val event = step.events(store).shareIn(scope, SharingStarted.Lazily)
+
+    // To ensure that all subscribers see SSE we wait until all known subscribers are active.
+    // This has to be updated every time we add another subscriber to the event.
+    private val threshold = if (response.dataChannelId == -1) 5 else 6
+    private val event =
+        step.events(store).shareIn(scope, SharingStarted.WhileSubscribedAtLeast(threshold))
```
tpettersen, line 73: I understand the issue and your solution. I just wonder if you have explored having some sort of cache with all events? The idea is that the subscribers can get all events the first time they subscribe and construct the state they need, before getting updates from on the incoming events. We got the same issue with presentation in a directMedia call on call start

drymarau: I don't think there's a good number for replay cache that fits our use case. Is it 1? 8? 64? More?

End goal is to remove dependency on SSE for consumers entirely and instead expose the current state. 

Some events may stay (e.g. FECC) in some form.

drymarau:
> I am looking forward for the state instead of the SSEs though, to avoid counting subscribers

Subscriber counting will still happen internally.

PR #335, opened 2024-02-26
feat(core): introduce sdk-core, retry function

drymarau:
This will wrap most of the calls to API that may throw `IOException` and retry such calls. This should hopefully help with network transitions as seen in #323.

This PR also introduces a new module, `:sdk-core`, which will contain things that may apply to a larger scope than just Infinity.


sdk-core/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Pexip SDK core
+
+## Using in your projects
+
+### Gradle
+
+```kotlin
+dependencies {
+    implementation("com.pexip.sdk:sdk-core:0.14.1")
```
tpettersen, line 11: Will it be available in 0.14.1?

drymarau: No, but easier to search and replace before next release.

drymarau:
> It looks ok 👍 Just a question before approving: It could be an idear to move `retry` into the `.await() `function itself perhaps - maybe as a configurable default param? It would then be available by default for all network calls we do, on all clients.

I considered this approach, but it has several issues:
1. We'll need a new API on Call to clone it done you can't retry an old Call
2. You won't be able to update parameters of the call (e.g. token) if they change between retries

tpettersen (approved):

PR #338, opened 2024-02-27
feat(conference): add a way to transform layout

drymarau:
This also renames `host_layout` to `layout` (since they are aliases) to better indicate that `host_layout`/`layout` alter the layout for guests in non-Virtual Auditorium calls.

### DIFF
--- a/sdk-conference/src/test/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImplTest.kt
+++ b/sdk-conference/src/test/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImplTest.kt
@@ -18,10 +18,13 @@ package com.pexip.sdk.conference.infinity.internal
 import app.cash.turbine.test
 import assertk.Assert
 import assertk.all
+import assertk.assertFailure
 import assertk.assertThat
 import assertk.assertions.containsExactly
 import assertk.assertions.extracting
+import assertk.assertions.hasCause
 import assertk.assertions.isEqualTo
+import assertk.assertions.isInstanceOf
 import assertk.assertions.isNotNull
 import assertk.assertions.isNull
 import assertk.assertions.key
@@ -35,10 +38,12 @@ import com.pexip.sdk.api.infinity.LayoutEvent
 import com.pexip.sdk.api.infinity.SplashScreenEvent
 import com.pexip.sdk.api.infinity.SplashScreenResponse
 import com.pexip.sdk.api.infinity.TokenStore
+import com.pexip.sdk.api.infinity.TransformLayoutRequest
 import com.pexip.sdk.conference.Element
 import com.pexip.sdk.conference.Layout
 import com.pexip.sdk.conference.LayoutId
 import com.pexip.sdk.conference.SplashScreen
+import com.pexip.sdk.conference.TransformLayoutException
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.test.runTest
 import kotlin.random.Random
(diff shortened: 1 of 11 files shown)


sdk-conference/src/test/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImplTest.kt
```diff
@@ -197,6 +202,73 @@ class ThemeImplTest {
...
+        assertFailure { theme.transformLayout(layout, guestLayout, enableOverlayText) }
+            .isInstanceOf<TransformLayoutException>()
+            .hasCause(t)
+    }
+
+    @Test
+    fun `transformLayout() succeeds`() = runTest {
+        val layout = Random.maybe { nextLayoutId() }
+        val guestLayout = Random.maybe { nextLayoutId() }
+        val enableOverlayText = Random.maybe { nextBoolean() }
```
tpettersen, line 241: it is a bit hard to have reliable tests with `Random.maybe`. If there is a bug in the `transformLayout` where for instance `enableOverlayText` is set to `null` no matter what, this test will pass if `Random.maybe` returns null. Next run it might fail, or not. For things that is careface if is true/false/null and is just an input value, this is perfectly fine ofc.

tpettersen (changes requested):
Looks good 👍 I think we just need to adjust the tests a bit

sdk-conference/src/test/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImplTest.kt
drymarau: There may be quite a few potential combinations of all the parameters, so `Random.maybe` helps a bit with reducing the boilerplate.

We usually have a dedicated test for when inputs actually matter, so it's a balancing act.

tpettersen: But inputs do matter in this test, and the test will not catch bugs created in function it tests. There is no other tests for `transformLayout` in `ThemeImpl` afaik?

tpettersen: Yes I understand that it is somewhat more cumbersome to test all inputs, but as it stands it can become flaky. IMO we should test the inputs separately to ensure we catch stuff. Maybe write separate tests or some parameterized thing?

drymarau: Again, any combination of inputs is valid since the validation is done server-side. We already check that the request matches inputs.

tpettersen: Does not matter if the validation is server-side if I send `enableOverlayText = true` to `transformLayout`, and `enableOverlayText = false` is passed along to the server.

drymarau: It is tested

tpettersen: change line 79 in `transformLayout` to `enableOverlayText = false`, and the test might perfectly well still pass. That is a uncaught bug

drymarau: Given that's a 50/50 chance, it will be caught on subsequent test runs.

tpettersen: = flaky test. And it is 33%, since it is nullable. 

We are probably not going to agree on this. I will approve it against my better judgement, and you can merge if you like

tpettersen (approved):

PR #341, opened 2024-02-29
fix(conference)!: make Participant.startTime nullable

drymarau:
There are some cases when it can be `null` in the payload, namely when dialing a user on Pexip Service. This may potentially be true other dialing scenarios.


tpettersen (approved):
hm. This is unfortunate, as the docs for the API says it should not be nullable. Hopefully it gets fixed in Infinity down the line

drymarau:
> hm. This is unfortunate, as the docs for the API says it should not be nullable.

I believe REST API docs claim no such thing (but they also omit the fact that it can be `null`).

PR #343, opened 2024-03-01
feat(conference): add Roster.me

drymarau:
This points to *your* participant object in the conference and will be useful for getting things like current "hand" state (whether raised or not), etc.

### DIFF
--- a/sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Roster.kt
+++ b/sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Roster.kt
@@ -29,6 +29,11 @@ public interface Roster {
     public val participants: StateFlow<List<Participant>>
 
     /**
+     * A [StateFlow] that represents *you* as the participant of this conference.
+     */
+    public val me: StateFlow<Participant?>
+
+    /**
      * Raises hand of the specified participant or self.
      *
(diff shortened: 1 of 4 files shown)


sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Roster.kt
```diff
@@ -28,6 +28,11 @@ public interface Roster {
      */
     public val participants: StateFlow<List<Participant>>
 
+    /**
+     * A [StateFlow] that represents *you* as the participant of this conference.
+     */
+    public val me: StateFlow<Participant?>
```
tpettersen, line 34: Optimally this should not be null. When the conference is started and you get a roster, that will at least include `me` - as I am there for sure. That will however require somewhat refactoring to achieve, passing along the first values after sync to `RosterImpl` removing encapsualtion and so on. A nullable Participant will have to do for now

tpettersen (approved):

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Roster.kt
drymarau: Given the async nature of how participant list is build it's not possible to achieve. A fake participant is doable, but its usefulness is limited.

PR #349, opened 2024-03-06
fix(conference): use debounce for event subscription

drymarau:
Prior approach still had some issues with `layout` event and API participants.


sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/InfinityConference.kt
```diff
@@ -67,11 +68,10 @@ public class InfinityConference private constructor(
...
     private val store = TokenStore.create(response)
 
-    // To ensure that all subscribers see SSE we wait until all known subscribers are active.
-    // This has to be updated every time we add another subscriber to the event.
-    private val threshold = if (response.dataChannelId == -1) 6 else 7
-    private val event =
-        step.events(store).shareIn(scope, SharingStarted.WhileSubscribedAtLeast(threshold))
+    private val event = step.events(store).shareIn(
+        scope = scope,
+        started = SharingStarted.WhileSubscribedWithDebounce(100.milliseconds),
```
tpettersen, line 73: Could be a constant

tpettersen (approved):
Seems to work good 👍

PR #356, opened 2024-03-12
feat(conference): add mute and unmute to Roster

drymarau:
Additionally, make every property and method in `Roster` have a default implementation that throws `NotImplementedError` to ensure binary compatibility for the consumers and to avoid breaking them (mostly their tests, really) on version upgrades.


tpettersen (approved):
Looks good 👌 At one point the sample app would benefit from catching up with all the new stuff in Roster and alike

PR #357, opened 2024-03-12
refactor(api): use default interface methods in InfinityService.kt

drymarau:
Additionally, deprecate all methods that accept `String` instead of a `Token`.


tpettersen (approved):
Looks good! Could have been an opportunity to bump the deprecation level to Error for all the `token: String` functions and use only the Token variants.

drymarau:
> Could have been an opportunity to bump the deprecation level to Error for all the `token: String` functions and use only the Token variants.

Sure, will do it.

PR #360, opened 2024-03-13
feat(conference): add spotlight and unspotlight to Roster

drymarau:


sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Roster.kt
```diff
@@ -91,6 +91,22 @@ public interface Roster {
...
+     */
+    public suspend fun spotlight(participantId: UUID? = null): Unit = throw NotImplementedError()
+
+    /**
+     * Disables the "spotlight" on a participant or self.
+     *
+     * @param participantId an ID of the participant, null for self
+     * @throws UnspotlightException if the operation failed
+     */
+    public suspend fun unspotlight(participantId: UUID? = null): Unit = throw NotImplementedError()
```
tpettersen, line 108: is unspotlight a valid word? Only reference I can find is in Zoom libs, other than that it is not in any dictionary afaik. Perhaps rename to `removeSpotlight`, `disableSpotlight` or similar?

sdk-conference/src/test/kotlin/com/pexip/sdk/conference/infinity/internal/RosterImplTest.kt
```diff
@@ -881,6 +883,212 @@ class RosterImplTest {
...
+            event = event,
+            participantId = participantId,
+            store = store,
+            step = object : InfinityService.ConferenceStep {},
+        )
+        roster.unspotlight(UUID.randomUUID())
+    }
+
+    @Test
+    fun `unspotlight() throws SpotlightException`() = runTest {
```
tpettersen, line 1002:
```suggestion
    fun `unspotlight() throws UnspotlightException`() = runTest {
```

tpettersen (approved):
Looks ok 👍 Just unsure of the naming

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Roster.kt
drymarau: Prefer to keep it short, as you mentioned there's at least one other library that uses it for a similar purpose so it should be OK.

PR #364, opened 2024-03-18
feat(conference): add a way to lock/unlock the conference

drymarau:


sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/Event.kt
```diff
@@ -126,6 +136,7 @@ public data class IncomingCancelledEvent(val token: String) : Event
 
 @Suppress("ktlint:standard:function-naming")
 internal fun Event(json: Json, id: String?, type: String?, data: String): Event? = when (type) {
+    "conference_update" -> json.decodeFromString<ConferenceUpdateEvent>(data).also(::println)
```
tpettersen, line 139:
```suggestion
    "conference_update" -> json.decodeFromString<ConferenceUpdateEvent>(data)
```

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Roster.kt
```diff
@@ -43,6 +43,15 @@ public interface Roster {
     public val presenter: StateFlow<Participant?>
         get() = throw NotImplementedError()
 
+    /**
+     * A [StateFlow] that represents whether the conference is locked.
+     *
+     * When a conference is locked, participants waiting to join are held at
+     * a "Waiting for Host" screen. These settings are only available to conference hosts.
+     */
+    public val locked: StateFlow<Boolean>
```
tpettersen, line 52: Feels off that the conference status belongs to Roster. It should move to the `Conference` at its root or wrap it in a separate class called `Status` or similar. In the clients lock/unlock will not live inside the Roster either, but in a meetings settings.

tpettersen (changes requested):
Lock/unlock should not be a part of Roster. Other than that it looks good :)

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Roster.kt
drymarau: Given that lock/unlock only affects the participant list I believe `Roster` is the best place for it, especially given that `admit` also lives in `Roster.

tpettersen: I get your point. But from a client perspective, locking and unlocking will not happen in the roster or participant list, rather from a meetings settings or similar page. `admit` is per participant, and makes sense to have in the Roster. `lock/unlock` is per conference.  Conference controls not related to participants should live in the `Conference` or a separate place IMO

drymarau: This conference control is related to participants, as I mentioned earlier.

tpettersen: I disagree. This API is about the conference, not participants. It is even located under conference controls in the API docs. We have a unlock api for participants which is of course natural to include in the roster.

drymarau: By that logic, `disconnectAll`, `lowerAllHands`, `muteAllGuests`, `unmuteAllGuests` and a few others should be under `Conference` as well. The shape of the REST API does not dictate how SDK API should look.

There's no concept of `Roster` in the API itself. It's just a useful construct to group things that affect participants (that includes locking/unlocking the list, as well as "unlocking" any specific participant).

tpettersen: I still disagree. The mentioned apis is affecting participants, but locking the conference is akin to `meeting layout` or `show name labels`. If a client is to create conference controls, it needs to pass the roster with participants and `me` attached to it, which does not belong there. 

But we have reached an impass. IMO this is wrong, and you put resolved + enabled auto-merge on this before agreeing on something. So the clients will have to live with or love this decision. It can be revisited if there are complaints down the road.

tpettersen (approved):

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Roster.kt
drymarau: In my opinion, UI doesn't drive API decisions such as this one - there are too many possibilities here where a consumer may want to put one control or the other (e.g. this is hidden in webapp3 under _Meeting controls_, but MMV for Android has a prominent lock/unlock button in the top app bar in the participant list screen).

You mentioned meeting layout and showing name labels. We have a place for this as well which is not `Conference` - `Theme`.

PR #366, opened 2024-03-19
feat(conference): add Participant.speaking property

drymarau:
This should indicate if the participant is currently talking.

Note that instead of exposing a _Stage_ we just add it as a property to ease the consumption of this API by consumers.


tpettersen (approved):
Seems good 👍

PR #375, opened 2024-03-25
release: 0.15.0

drymarau:


tpettersen (approved):
I looked through the commits from 0.14.0 until now, and it seems you caught them all 🙌  

If there was time, I would have wished for the sample app to include all the changes, but I guess it will have to wait

drymarau:
> If there was time, I would have wished for the sample app to include all the changes, but I guess it will have to wait

Agree, we'll need to get back to it sooner rather than later.

PR #378, opened 2024-03-26
refactor(api,conference)!: replace java.util.UUID with value classes

drymarau:
Note that in the normal use this will only break tests that instantiate classes or call functions that use `java.util.UUID`.


sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Messenger.kt
```diff
@@ -63,7 +67,12 @@ public interface Messenger {
         message = "Use the coroutines version of this method.",
         level = DeprecationLevel.ERROR,
     )
-    public fun send(participantId: UUID, type: String, payload: String, callback: SendCallback)
+    public fun send(
```
tpettersen, line 70: Since this is deprecated with ERROR, we could as well just remove the method completely.

sdk-conference/src/test/kotlin/com/pexip/sdk/conference/infinity/internal/RosterImplTest.kt
```diff
@@ -79,13 +80,14 @@ import com.pexip.sdk.api.infinity.ServiceType as ApiServiceType
...
-    private lateinit var participantId: UUID
     private lateinit var store: TokenStore
 
+    private var participantId: ParticipantId by Delegates.notNull()
+
     @BeforeTest
     fun setUp() {
         event = MutableSharedFlow(extraBufferCapacity = 1)
-        participantId = UUID.randomUUID()
+        participantId = ParticipantId(Random.nextString(8))
```
tpettersen, line 90: convenience-nit: could be nice with a utility for `Random.nextParticipant()` instead of using `ParticipantId(Random.nextString(8))` everywhere

tpettersen (approved):

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Messenger.kt
drymarau: Sure, it will be removed before the next release.

PR #380, opened 2024-03-27
refactor(core,infinity): convert modules to multiplatform

drymarau:


sdk-core/src/jvmMain/kotlin/com/pexip/sdk/core/internal/IOException.jvm.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.pexip.sdk.core.internal
+
+@Suppress("ACTUAL_WITHOUT_EXPECT")
+internal actual typealias IOException = java.io.IOException
```
tpettersen, line 19: I am getting a warning on this and the `expect` that says:
```
expect'/'actual' classes (including interfaces, objects, annotations, enums, and 'actual' typealiases) are in Beta. You can use -Xexpect-actual-classes flag to suppress this warning. Also see: https://youtrack.jetbrains.com/issue/KT-61573
```

Should we enable the flag or suppress the warning?

tpettersen (approved):
Works good when testing

PR #401, opened 2024-04-30
feat(infinity): introduce Node

drymarau:
This PR adds a `Node` that represents the address of the Infinity node
that the SDK will communicate with.

Note that while the `Node` includes `port`, in real world use it will
almost exclusively be 443 since Infinity forces you to use TLS over said
port. It's exposed here to allow testing against `MockWebServer`.

We also added a suspending `NodeResolver` that will return a `Node`
list and deprecated the one found in `:sdk-api` module.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -5,6 +5,7 @@
 /assetWizardSettings.xml
 /compiler.xml
 /deploymentTargetDropDown.xml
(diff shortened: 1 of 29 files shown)


tpettersen (approved):
Works good when testing 👍 The deprecations could for the sake of ease point to which module/function you should use instead

PR #406, opened 2024-05-01
refactor(api)!: simplify TokenStore

drymarau:
This PR replaces `AtomicReference` as the in-memory store of the `Token`
to a `MutableStateFlow`, giving us a way to observe the changes in the
`TokenStore`.

Notably, it's now an internal API (and was never meant to be used
outside of the project anyway). Consecutively, it's no longer an
interface, but a concrete class.


sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/TokenStore.kt
```diff
@@ -60,6 +105,11 @@ public interface TokenStore {
          * @param releaseToken a function that releases the [Token]
          * @param onFailure a callback to notify callers about failures
          */
+        @Suppress("EXTENSION_SHADOWED_BY_MEMBER")
```
tpettersen, line 108: This should just be removed as it is not possible to reach. Using `store.refreshTokenIn(..)` will point to the new method

tpettersen (approved):

sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/TokenStore.kt
drymarau: Reason for it to stay is the import that may still be present if consumers used that method.

We'll remove it after the next release.

PR #411, opened 2024-05-08
feat(api): add call_tag support

drymarau:


sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/RequestTokenResponse.kt
```diff
@@ -29,31 +29,33 @@ public data class RequestTokenResponse(
...
+    val dataChannelId: Int = -1,
     @Serializable(with = DurationAsMillisecondsSerializer::class)
     @SerialName("client_stats_update_interval")
-    public val clientStatsUpdateInterval: Duration = Duration.INFINITE,
+    val clientStatsUpdateInterval: Duration = Duration.INFINITE,
     @Transient
-    public val directMediaRequested: Boolean = false,
+    val directMediaRequested: Boolean = false,
+    @SerialName("call_tag")
+    val callTag: String = "",
```
tpettersen, line 60: Smallest of nits - I generally like to keep the transient fields at the bottom, so it is easier to keep track of what is what.

tpettersen (commented):
Looks good! Before approving, I am just wondering if we should add `call_tag` support for `ParticipantCreate` as well to make it complete?

tpettersen (approved):

PR #419, opened 2024-05-17
feat(conference): add avatar.jpg URL

drymarau:
This can be used to provide a custom icon for a specific conference
alias.

### DIFF
--- a/sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Theme.kt
+++ b/sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Theme.kt
@@ -24,14 +24,22 @@ import kotlinx.coroutines.flow.StateFlow
 public interface Theme {
 
     /**
+     * A [StateFlow] of the logo URL
+     */
+    public val avatar: StateFlow<String>
+        get() = throw NotImplementedError()
+
+    /**
      * A [StateFlow] of current [Layout] state.
      */
     public val layout: StateFlow<Layout?>
+        get() = throw NotImplementedError()
 
     /**
      * A [StateFlow] of [SplashScreen]s that should be displayed.
      */
     public val splashScreen: StateFlow<SplashScreen?>
+        get() = throw NotImplementedError()
 
     /**
      * Transforms the current conference layout given sufficient privileges.
@@ -50,5 +58,7 @@ public interface Theme {
         layout: LayoutId? = null,
         guestLayout: LayoutId? = null,
         enableOverlayText: Boolean? = null,
-    )
+    ) {
+        throw NotImplementedError()
+    }
 }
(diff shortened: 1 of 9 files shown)


sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImpl.kt
```diff
@@ -121,20 +125,21 @@ internal class ThemeImpl(
         is ElementResponse.Unknown -> null
     }
 
-    private fun InfinityService.ConferenceStep.availableLayouts(store: TokenStore) = store.asFlow()
+    private fun InfinityService.ConferenceStep.availableLayouts(store: TokenStore) = store.token
+        .take(1)
```
tpettersen, line 129: Why these `take(1)`s? Do we have more than 1 token?

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Theme.kt
```diff
@@ -23,15 +23,23 @@ import kotlinx.coroutines.flow.StateFlow
  */
 public interface Theme {
 
+    /**
+     * A [StateFlow] of the logo URL
+     */
+    public val avatar: StateFlow<String>
+        get() = throw NotImplementedError()
+
```
tpettersen, line 31: According to the API docs, it says 
```
Obtains the image to display to represent a conference participant or directory contact.
```

Is this right to put as a "logo" in the theme?

tpettersen (commented):
A couple of questions before approving

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Theme.kt
drymarau: You're looking at the avatar of a participant, but this is for a conference itself. I believe it's not documented.

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImpl.kt
drymarau: It's a StateFlow so there will be emissions when the token refreshes. But the layouts don't need to be updated every minute, just once (or if we encounter an error)

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Theme.kt
tpettersen: 😭 I just love undocumented APIs

tpettersen (approved):

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/Theme.kt
drymarau: It's like a treasure hunt!

PR #421, opened 2024-05-23
feat(api,conference): respect parent_participant_uuid

drymarau:
In certain situations a `RequestTokenResponse` may contain a
`parent_participant_uuid` field which is effectively the participant we'd
like to represent and make calls on behalf of.


drymarau:
@tpettersen, please note that this can be triggered through SIP Session ID flow and will require some creative setup for testing.

tpettersen (approved):

PR #427, opened 2024-06-03
fix(conference): emit a fake LayoutEvent

drymarau:
Several versions of Infinity have a bug where `layout` event will not be
emitted under certain circumstances (e.g. joining using SIP Session ID,
API-only participant, etc.).

This workaround emits a fake `layout` event without much set. During
normal operation it will be followed by a real one, overriding the
values. For SIP Session ID flow it will let the `Theme` construct the
`Layout`, albeit in a limited fashion.

### DIFF
--- a/sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImpl.kt
+++ b/sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImpl.kt
@@ -42,6 +42,7 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.filterIsInstance
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.retryWhen
 import kotlinx.coroutines.flow.stateIn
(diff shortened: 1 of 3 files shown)


sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImpl.kt
```diff
@@ -60,7 +61,10 @@ internal class ThemeImpl(
         .stateIn(scope, SharingStarted.Eagerly, step.avatar(store.token.value))
 
     override val layout: StateFlow<Layout?> = combine(
-        flow = event.filterIsInstance<LayoutEvent>(),
+        flow = flow {
+            emit(LayoutEvent(LayoutId("unknown")))
```
vadymmarkov, line 65: Is this an actual layout id?

vadymmarkov (approved):

sdk-conference/src/main/kotlin/com/pexip/sdk/conference/infinity/internal/ThemeImpl.kt
drymarau: No, but given that custom layouts are a thing it can be. Maybe `""` would be a better fit?

vadymmarkov: `""` is probably safer in that case.

PR #448, opened 2024-07-30
feat(api): add client_mute, client_unmute and is_client_muted

drymarau:
A new API introduced in v36 that allows a client to signal whether
they're locally muted.


vadymmarkov (approved):
Looks good.

PR #456, opened 2024-08-12
fix(api): handle nanos and millis better

drymarau:
Serializers previously didn't handle duration values that included a
fraction well. This should improve things a bit and not make the
application crash when on v36.

### DIFF
--- a/sdk-api/src/test/kotlin/com/pexip/sdk/api/infinity/internal/RequestTokenResponseSerializerTest.kt
+++ b/sdk-api/src/test/kotlin/com/pexip/sdk/api/infinity/internal/RequestTokenResponseSerializerTest.kt
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2024 Pexip AS
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


sdk-api/src/test/kotlin/com/pexip/sdk/api/infinity/internal/RequestTokenResponseSerializerTest.kt
```diff
@@ -0,0 +1,93 @@
...
+                    version = VersionResponse(
+                        versionId = "36",
+                        pseudoVersion = "79019.0.0",
+                    ),
+                    chatEnabled = true,
+                    guestsCanPresent = true,
+                    serviceType = ServiceType.UNKNOWN,
+                    directMedia = true,
+                    dataChannelId = 3,
+                    clientStatsUpdateInterval = 4411.068996234768.milliseconds,
```
vadymmarkov, line 84: Is this `clientStatsUpdateInterval` actually something we need to parse?

vadymmarkov (approved):

sdk-api/src/test/kotlin/com/pexip/sdk/api/infinity/internal/RequestTokenResponseSerializerTest.kt
drymarau: It's supposed to be used during direct media to send media stats to Infinity, but we haven't got around to implementing that yet.

PR #459, opened 2024-08-15
feat(conference): add clientMute and clientUnmute to Roster

drymarau:
This is only really useful when the media is provided by some other
means (e.g. SIP) and we want to sync local mute state with Infinity.


vadymmarkov (approved):
Looks good.

PR #468, opened 2024-09-05
feat(conference): use parentParticipantId from v35.1

drymarau:
Additionally, use `parentParticipantId` for `mute` and `unmute` at all times.


vadymmarkov (approved):
The code looks good. Only one question, what is a parent participant? 😄

drymarau:
> Only one question, what is a parent participant? 😄

Parent participant is usually a SIP device that receives a SIP Session ID. SIP Session ID can be used as an alternative to a conference alias. Upon joining, the participant created becomes a child of the SIP participant with the ability to control the parent. Note that the child participant will not be shown in the roster list, only the parent will be.

You can check if you're a child participant by looking for `parent_participant_uuid` in the `request_token` response. When using the soft client flow, you will not see this property.

PR #490, opened 2024-10-04
feat: set jvm toolchain to 17

drymarau:
This is mostly brought up by updates to WebRTC (required JDK17) and has
been generally supported by Android for some time now.


drymarau:
AGP appears to be coming with its own Dokka plugin (outdated) and has some issues when generating Javadoc JAR on JDK17.

This PR disables the built-in Javadoc JAR generation and instead wires Dokka output to the achieve the same effect.

vadymmarkov (approved):

PR #498, opened 2024-10-11
feat(api): add BreakoutStep

drymarau:
This PR includes basic building blocks to implement breakout feature,
including:
- `BreakoutId`, which is unique to a breakout
- `BreakoutStep`, where most of the breakout-related methods will reside

### DIFF
--- a/sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
+++ b/sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
@@ -20,6 +20,7 @@ package com.pexip.sdk.api.infinity
 import com.pexip.sdk.api.Call
 import com.pexip.sdk.api.EventSourceFactory
 import com.pexip.sdk.api.infinity.internal.InfinityServiceImpl
+import com.pexip.sdk.infinity.BreakoutId
 import com.pexip.sdk.infinity.CallId
 import com.pexip.sdk.infinity.LayoutId
 import com.pexip.sdk.infinity.Node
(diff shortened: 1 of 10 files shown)


sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
```diff
@@ -332,6 +340,24 @@ public interface InfinityService {
...
+        /**
+         * The conference step that produced this breakout step.
+         */
+        public val conferenceStep: ConferenceStep
+            get() = throw NotImplementedError()
+
+        /**
+         * A breakout ID that this breakout step will use.
+         */
+        public val breakoutId: BreakoutId
```
larswdahl, line 357: Maybe `breakoutRoomId` would be a more descriptive name for this property? I was almost sure that it was the room id, but I had to double check 🙂

larswdahl (approved):

sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/InfinityService.kt
drymarau: I believe `breakout` can be used as a [noun](https://www.merriam-webster.com/dictionary/breakout). This is also shown in the REST API (e.g. `breakouts/:breakoutId` in the path).

Just prefer to keep it shorter if possible without losing too much context 🙂

larswdahl: Sounds good to me. Let's keep it as it is to be consistent with the API 🙂

PR #511, opened 2024-10-27
feat(media): use software ns and aec

drymarau:
There has been a number of reports of hardware NS and AEC not working on
a range of devices.

Instead, we'll use software NS and AEC provided by WebRTC and consumers
can customize this behavior by providing a custom `AudioDeviceModule` if
they wish to use hardware implementation.


drymarau:
> Could it be the other way around, so consumers disable hw ns and aec themselves?

Possible, but the intent is broad compatibility and software NS & AEC gives that while hardware is a bit fickler and best served by an allowlist of devices on the consumer side.

vadymmarkov (approved):

PR #526, opened 2024-11-08
feat(media): add VideoTrackEmbeddedRenderer

drymarau:
This is a TextureView-based Composable that is embedded directly in the
UI hierarchy.

Unlike `VideoTrackRenderer` effects such as clipping can be safely
applied to this Composable (see the modified self-view for an example of
this).

### DIFF
--- a/sdk-media-webrtc-compose/api/sdk-media-webrtc-compose.api
+++ b/sdk-media-webrtc-compose/api/sdk-media-webrtc-compose.api
@@ -22,6 +22,10 @@ public final class com/pexip/sdk/media/webrtc/compose/FrameResolution {
 	public fun toString ()Ljava/lang/String;
 }
 
+public final class com/pexip/sdk/media/webrtc/compose/VideoTrackEmbeddedRendererKt {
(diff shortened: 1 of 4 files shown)


sdk-media-webrtc-compose/api/sdk-media-webrtc-compose.api
```diff
@@ -22,6 +22,10 @@ public final class com/pexip/sdk/media/webrtc/compose/FrameResolution {
 	public fun toString ()Ljava/lang/String;
 }
 
+public final class com/pexip/sdk/media/webrtc/compose/VideoTrackEmbeddedRendererKt {
```
vadymmarkov, line 25: Noob question - is this generated?

vadymmarkov (approved):

sdk-media-webrtc-compose/api/sdk-media-webrtc-compose.api
drymarau: Yes, anything that is public API will be added to `.api` files for binary compatibility validation.

PR #527, opened 2024-11-08
feat(api): set breakout_capable to false by default

drymarau:
This means that consumers should explicitly opt-in to support breakouts
feature. By default, this is not supported and the behavior matches that
of a guest on Infinity v36+.

Note that on v34 and v35 if the participant is being transfered to a
breakout they may not receive `refer` event, which means that they will
stay in the main room.

### DIFF
--- a/sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/RequestTokenRequest.kt
+++ b/sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/RequestTokenRequest.kt
@@ -15,9 +15,12 @@
  */
 package com.pexip.sdk.api.infinity
 
+import kotlinx.serialization.EncodeDefault
+import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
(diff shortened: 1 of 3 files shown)


sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/RequestTokenRequest.kt
```diff
@@ -36,4 +39,7 @@ public data class RequestTokenRequest(
     val directMedia: Boolean = false,
     @SerialName("call_tag")
     val callTag: String = "",
+    @EncodeDefault
+    @SerialName("breakout_capable")
```
vadymmarkov, line 43: Did it turn out that this field was already in REST API?

vadymmarkov (approved):

sdk-api/src/main/kotlin/com/pexip/sdk/api/infinity/RequestTokenRequest.kt
drymarau: Yes, starting from v34. There was a bug that prevented `refer` event though, but this will be fixed in v36.

